PR #3, opened 2018-07-24
delete default compute engine service account stability fix

ryanckoch:
Uses triggers to account for more cases of existing state of already created projects.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -104,8 +104,7 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
   Default compute service account retrieval
  *****************************************/
 data "google_compute_default_service_account" "default" {
-  project    = "${local.project_id}"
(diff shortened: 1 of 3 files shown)


main.tf
```diff
@@ -104,19 +104,23 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
   Default compute service account retrieval
  *****************************************/
 data "google_compute_default_service_account" "default" {
-  project    = "${local.project_id}"
-  depends_on = ["google_project_service.project_services"]
```
morgante: @ryanckoch Why is this dependency removed? The default compute service account won't appear until the compute API is activated.

ryanckoch: Because of an issue where it causes the dependent trigger to run on every apply:
https://github.com/hashicorp/terraform/issues/11806

PR #4, opened 2018-07-25
Adding new App Engine support

ryanckoch:
See https://github.com/terraform-providers/terraform-provider-google/pull/1503 for context.

Merge after: https://github.com/terraform-google-modules/terraform-google-project-factory/pull/3

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -128,23 +128,6 @@ resource "null_resource" "delete_default_compute_service_account" {
 }
 
 /******************************************
-  Default app engine service account deletion
- *****************************************/
-resource "null_resource" "delete_default_app_engine_service_account" {
-  count = "${local.app_engine_enabled ? 1 : 0}"
-
-  provisioner "local-exec" {
-    command = "${path.module}/scripts/delete-service-account.sh ${local.project_id} ${var.credentials_path} ${local.project_id}@appspot.gserviceaccount.com"
-  }
-
-  triggers {
-    app_engine_enabled = "${local.app_engine_enabled}"
-  }
-
-  depends_on = ["google_project_service.project_services"]
-}
-
-/******************************************
   Default Service Account configuration
  *****************************************/
 resource "google_service_account" "default_service_account" {
--- a/main.tf
+++ b/main.tf
@@ -37,6 +37,12 @@ locals {
   project_bucket_name = "${var.bucket_name != "" ? var.bucket_name : format("%s-state", var.name)}"
   create_bucket       = "${var.bucket_project != "" ? "true" : "false"}"
   gsuite_group        = "${var.group_name != "" || var.create_group}"
+  app_engine_enabled  = "${length(keys(var.app_engine)) > 0 ? true : false}"
+
+  app_engine_config = {
+    enabled  = "${list(var.app_engine)}"
+    disabled = "${list()}"
+  }
 }
 
 /******************************************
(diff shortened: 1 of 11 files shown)


morgante:
@ryanckoch Can you please add a test to check that an app is actually created on the project?

ryanckoch:
@morgante Done

main.tf
```diff
@@ -37,6 +37,12 @@ locals {
   project_bucket_name = "${var.bucket_name != "" ? var.bucket_name : format("%s-state", var.name)}"
   create_bucket       = "${var.bucket_project != "" ? "true" : "false"}"
   gsuite_group        = "${var.group_name != "" || var.create_group}"
+  app_engine_enabled  = "${length(keys(var.app_engine)) > 0 ? true : false}"
+
+  app_engine_config = {
+    enabled  = "${list(var.app_engine)}"
```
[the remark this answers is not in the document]
ryanckoch: It gets the length of keys in app_engine which is a list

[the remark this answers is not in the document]
morgante: But then I'm confused about how line 86 works? https://github.com/terraform-google-modules/terraform-google-project-factory/pull/4/files/a51eb1f65e51143af5b7dff92d9ce2fa268cfb38#diff-7a370d8342e7203b805911c92454f0f4R86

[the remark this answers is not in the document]
ryanckoch: local.app_engine_config is a map of lists with keys `enabled` and `disabled`. Line 86 is referencing that list by key depending if local.app_engine_enabled is true or false. If local.app_engine_enabled is true, it sets `app_engine` to the `enabled` list which is the users config, otherwise it uses the `disabled` list which is empty. It is a dirty hack, but the only way I could see to pass a completely empty list to `app_engine` and still be able to accept the `app_engine` variable as a map in our module. Otherwise, it would set `app_engine = [{}]` which isnt an empty list and tells the provider to setill enable app_engine.

main.tf
```diff
@@ -76,6 +82,8 @@ resource "google_project" "project" {
   auto_create_network = "${var.auto_create_network}"
 
   labels = "${var.labels}"
+
+  app_engine = "${local.app_engine_config["${local.app_engine_enabled ? "enabled" : "disabled"}"]}"
```
morgante: @ryanckoch This is hard to parse and somewhat confusing. 

Wouldn't `app_engine = "${local.app_engine_enabled ? var.app_engine : list()}"` work just as well?

ryanckoch: It wont allow for a list in a conditional. That would definitely be much cleaner, but that is the reason for the hack app_engine_config map.

morgante: How about `"${local.app_engine_enabled ? local.app_engine_config["enabled"] : local.app_engine_config["disabled"]}"`

ryanckoch: Also worth noting, that although the examples in the documentation show app_engine as accepting a map, it accepts a list with a max length of 1.
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_google_project.go#L95

Have created an issue on the provider: https://github.com/terraform-providers/terraform-provider-google/issues/1821

morgante:
Please address [this comment](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/4#discussion_r205880933) and improve the app engine integration test by using (a) a different region + (b) an additional non-default app_engine flag + verifying both are set corectly.

main.tf
ryanckoch: Same:
```
Error: module.project-factory.google_project.project: 1 error(s) occurred:

* module.project-factory.google_project.project: At column 3, line 1: conditional operator cannot be used with list values in:

${local.app_engine_enabled ? local.app_engine_config["enabled"] : local.app_engine_config["disabled"]}
```

ryanckoch:
@morgante The above comment has been addressed. It is worth noting that setting custom feature_settings requires multiple applies: https://github.com/terraform-providers/terraform-provider-google/issues/1826

PR #14, opened 2018-08-29
First pass at removing the gsuite requirement for project creation

CPearring:
Still todo: update readme files for both the 'core' module and the 'gsuite_enabled' module. 

Also updated some tests, however still need to refactor tests and documentation.


morgante:
@CPearring Can you make sure to resolve merge conflicts?

CPearring:
Oh man did that work?

modules/core_project_factory/variables.tf
```diff
@@ -0,0 +1,105 @@
...
+  description = "The ID of the billing account to associate this project with"
+}
+
+variable "folder_id" {
+  description = "The ID of a folder to host this project"
+  default     = ""
+}
+
+variable "group_name" {
+  description = "A group to control the project by being assigned group_role - defaults to ${project_name}-editors"
```
morgante: This will default to no group with the refactor.

modules/gsuite_enabled/main.tf
```diff
@@ -71,5 +84,27 @@ resource "gsuite_group_member" "api_s_account_api_sa_group_member" {
   email = "${local.api_s_account}"
   role  = "MEMBER"
 
-  depends_on = ["google_project_service.project_services"]
+  depends_on = ["module.project-factory"]
```
morgante: Can't depend on a module, should depend on an output of the module.

modules/gsuite_enabled/variables.tf
```diff
@@ -0,0 +1,120 @@
...
+  description = "A project to create a GCS bucket (bucket_name) in, useful for Terraform state (optional)"
+  default     = ""
+}
+
+variable "bucket_name" {
+  description = "A name for a GCS bucket to create (in the bucket_project project), useful for Terraform state (optional)"
+  default     = ""
+}
+
+variable "api_sa_group" {
```
morgante: We should move the G Suite variables into their own file.

test/integration/gcloud/integration.bats
```diff
@@ -17,7 +17,7 @@
 }
 
 @test "Terraform plan, ensure connection and creation of resources" {
-
+  skip
```
morgante: Why are we skipping these tests? Guessing that's a mis-commit.

CPearring:
ack on the fixes, should be out shortly

modules/core_project_factory/variables.tf
CPearring: true, should I make it default to ${project_name}-editors? or change the description to have it default to no group? My issue here being that if the default group doesn't exist, the apply will error out.

CPearring:
alright changes should be out.

modules/gsuite_enabled/main.tf
```diff
@@ -14,10 +14,25 @@
...
-If we don't need to work with gsuite, we can use the project-factory without installing/configuring gsuite,
-it is necessary to delete this file that contains the gsuite resources.
-*/
+/******************************************
+  Locals configuration
+ *****************************************/
+
+locals {
+  project_number    = "${module.project-factory.project_number}"
+  api_s_account     = "${format("%s@cloudservices.gserviceaccount.com", local.project_number)}"
```
morgante: Instead of re-calculating these outside the module, why not add them as outputs to the core module?

examples/group_project/main.tf
```diff
@@ -15,7 +15,8 @@
  */
 
 locals {
-  credentials_file_path = "${path.module}/sa-key.json"
+  // credentials_file_path = "${path.module}/sa-key.json"
```
morgante: Please remove commented code.

modules/gsuite_enabled/main.tf
CPearring: definitely should have done this earlier!

Issue #20, opened 2018-09-07
Organization viewer should not be needed


ocervell:
We should be able to create a project without having roles/organization.viewer.

Currently the org_id is used in 2 places:

- [x] `google_project` resource (not needed unless the project is straight under the org)
- [x] `google_organization` (this resource is used to get the org domain in order to construct the default service account email string. This can be optional by letting the user pass 'domain' input directly

PR #21, opened 2018-09-10
Make `org_id` and `roles/organization.viewer` role optional

ocervell:
Currently the `org_id` is used to get the domain name required to create / get a GSuite group named `<group_name>@<domain>`. 

The domain name is gotten using the `google_organization` resource that requires the `roles/organization.viewer` role. This resource should be activated only if `group_name` is passed. 

Currently when `group_name` is not passed, the `google_organization` resource is still called, and that will fail if for some reason one does not have the Organization Viewer role.

This PR makes `org_id` optional and adds support for a `domain` flag that will replace the it in the previous scenario, so that the project creation can still go on.


morgante:
LGTM

ocervell:
Fixes #20

PR #22, opened 2018-09-13
change README helper script to correct script and updated usage

reechar-goog:


morgante:
Thanks @reechar-goog!

PR #24, opened 2018-09-14
Work in progress: Manage multiple service accounts

adrienthebo:
This pull request adds support for managing multiple service accounts in a given project.

### API changes

```diff
variables:
+ service_accounts
+ impersonated_user_email
- sa_role
- sa_group

outputs:
- service_account_id
- service_account_email
- service_account_name
- service_account_display_name

resources:
- service account membership in "roles/storage.admin" on the project bucket
- G Suite group membership in "roles/iam.serviceAccountUser" on the default service account
```

### TODO

* [ ] Add kitchen-terraform tests
* [ ] Verify that changing the `service_accounts` variable doesn't mangle role deletion/recreation
* [ ] G Suite group access to service accounts
* [ ] Service accounts with network access have `roles/compute.networkUser` on the shared VPC project

This pull request is a work in progress and may be rebased if necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ module "project-factory" {
   billing_account     = "ABCDEF-ABCDEF-ABCDEF"
   group_role          = "roles/editor"
   shared_vpc          = "shared_vpc_host_name"
-  sa_group            = "test_sa_group@yourdomain.com"
   credentials_path    = "${local.credentials_file_path}"
 
   shared_vpc_subnets = [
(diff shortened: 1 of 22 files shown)


adrienthebo:
Tests have been added at https://github.com/terraform-google-modules/terraform-google-project-factory/pull/32, when that PR has been merged this will be rebased on top of it and appropriate tests added.

adrienthebo:
While working on this pull request I discovered a bug that comes out of using a script to manage IAM roles and groups. If a user deletes a role from the `service_accounts` map  and re-runs Terraform, Terraform will effectively "forget" that the service account was assigned to that role and will *not* remove the service account from the role. The same thing happens for groups. When Terraform invokes the null resource running the scripts it doesn't provide the diff between the old and new service account maps, so there's no way to determine which groups should actually be removed.

In summary, using a script to manage group and role membership appears to be fundamentally broken.

While Terraform is very noisy when updating multiple resources defined with a list and `count` that behavior is correct, while this approach seems inherently dangerous. because of this I'm closing this pull request.

PR #31, opened 2018-09-21
Refactor to remove G Suite dependency

morgante:


main.tf
```diff
@@ -14,315 +14,35 @@
...
-
-/*******************************************
-  Project creation
- *******************************************/
-resource "google_project" "project" {
+module "project-factory" {
+  source              = "modules/core_project_factory"
+  random_project_id   = "${var.random_project_id}"
+  org_id              = "${var.org_id}"
+  domain = "${var.domain}"
```
morgante: Please run `terraform fmt`.

modules/core_project_factory/main.tf
```diff
@@ -0,0 +1,322 @@
...
+  domain                 = "${var.domain != "" ? var.domain : var.org_id != "" ? join("", data.google_organization.org.*.domain) : ""}"
+  s_account_fmt          = "${format("serviceAccount:%s", google_service_account.default_service_account.email)}"
+  api_s_account          = "${format("%s@cloudservices.gserviceaccount.com", local.project_number)}"
+  api_s_account_fmt      = "${format("serviceAccount:%s", local.api_s_account)}"
+  gke_shared_vpc_enabled = "${var.shared_vpc != "" && contains(var.activate_apis, "container.googleapis.com") ? "true" : "false"}"
+  gke_s_account          = "${format("service-%s@container-engine-robot.iam.gserviceaccount.com", local.project_number)}"
+  gke_s_account_fmt      = "${local.gke_shared_vpc_enabled ? format("serviceAccount:%s", local.gke_s_account) : ""}"
+  project_bucket_name    = "${var.bucket_name != "" ? var.bucket_name : format("%s-state", var.name)}"
+  create_bucket          = "${var.bucket_project != "" ? "true" : "false"}"
+  gsuite_group           = "${var.group_name != "" || var.create_group}"
```
morgante: `create_group` is not part of the core project factory.

modules/gsuite_enabled/gsuite_variables.tf
```diff
@@ -0,0 +1,30 @@
+/**
```
adrienthebo: This should be renamed to `variables.tf` to conform with Terraform best practices.

modules/gsuite_enabled/main.tf
```diff
@@ -70,6 +83,28 @@ resource "gsuite_group_member" "api_s_account_api_sa_group_member" {
...
+module "project-factory" {
+  source              = "../core_project_factory"
+  random_project_id   = "${var.random_project_id}"
+  org_id              = "${var.org_id}"
+  domain = "${var.domain}"
+  name                = "${var.name}"
+  shared_vpc          = "${var.shared_vpc}"
+  billing_account     = "${var.billing_account}"
+  folder_id           = "${var.folder_id}"
+  group_name          = "${var.create_group ? "${gsuite_group.group.name}" : local.group_name}"
```
adrienthebo: The quotation and interpolation of `"${gsuite_group.group.name}"` looks to be redundant; since we're already in an interpolation expression we should be able to replace that with `gsuite_group.group.name`.

variables.tf
```diff
@@ -48,25 +48,15 @@ variable "folder_id" {
 }
 
 variable "group_name" {
-  description = "A group to control the project by being assigned group_role - defaults to ${project_name}-editors"
+  description = "A group to control the project by being assigned group_role"
   default     = ""
 }
 
-variable "create_group" {
```
adrienthebo: The deletion of these variables changes the outward facing APIs. Are these variables no longer available or does one have to use `modules/gsuite_enabled` in order to access these variables?

modules/core_project_factory/main.tf
```diff
@@ -0,0 +1,321 @@
...
+  app_engine_config = {
+    enabled  = "${list(var.app_engine)}"
+    disabled = "${list()}"
+  }
+}
+
+/******************************************
+  Group email to be used on resources
+ *****************************************/
+data "null_data_source" "data_final_group_email" {
```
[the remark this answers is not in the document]
CPearring: Great point, I totally missed this. But yes moving over to local can be done and it would be one less thing to have to move in the state file.

ocervell:
Can this be merged soon ? I'm with a customer and they don't want to pull in the GSuite module.

adrienthebo:
Master merged into gsuite-refactor, this is ready for merge.

ocervell:
@adrienthebo could you give an example how we would use this module with github URLs ?  

I understand the 'core module' should work just fine with the base url (https://github.com/terraform-google-modules/terraform-google-project-factory) because it's defaulted into the `main.tf`, but how would we pick the GSuite-enabled one ?

adrienthebo:
@ocervell I was able come up with this:

```hcl
module "gsuite-factory" {
  source            = "github.com/terraform-google-modules/terraform-google-project-factory//modules/gsuite_enabled?ref=gsuite-refactor"
  name              = "hello-world"
  random_project_id = true
  billing_account   = "000000-000000-000000"
  credentials_path  = "credentials.json"
}
```

A couple of notes about the source url:
- The `//` separator indicates that the following path components are a [subdirectory within the Git repository](https://www.terraform.io/docs/modules/sources.html#modules-in-package-sub-directories)
- If you need to use this branch immediately, keep the `?ref=gsuite-refactor` URI parameter; when we merge this PR you should be able to drop that.

Give that a shot and see if this works for you?

morgante:
Replaced with #73.

PR #32, opened 2018-09-24
Convert test coverage to test-kitchen and InSpec

adrienthebo:
This pull request converts our test coverage from BATS to Test Kitchen and InSpec. These tools offer more in the ways of test environment setup and teardown, provide more robust test coverage, and ease test driven development workflows.

This work is based on the work done by @ryanckoch (in https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/commit/f9de8c2eace3188e5eaafe341b90e5ef82a4e1d1), but deviates in a few areas that I'd like reviewed. The notable areas are as follows:

1. A test module has been extracted from the original runner script, instead of using the runner script to template out the Terraform configuration files.
1. The `test/integration/gcloud/run.sh` script has been refactored into a setup only script (at `test/integration/gcloud/setup.sh` that generates configuration files for Terraform and InSpec. It still relies on a `config.sh` script for setting up those files.
1. The Terraform test module is configured via a `terraform.tfvars` file, which is generated by `setup.sh`.
1. The InSpec test profile is configured by `inspec-attributes.yml` instead of relying on environment variables.
1. The InSpec test profile is located at `test/integration/default`.
1. The test-kitchen configuration file is located at the root of the project instead of `test/integration/gcloud/.kitchen.yml`.

I've tried to optimize this test setup for the developer to make it as easy as possible to re-run tests quickly during development. @ryanckoch since this diverges from the work that you've already done I want to make sure that these differences seem sensible and won't drive us crazy down the road.

I'm also quite familiar with RSpec and have used some idioms based on my past experience, so I'd like people to eyeball the tests to make sure they're sensible and not too magic, because magic is meant for circuses and not for integration tests.


adrienthebo:
@jmickle I'd love your review of these changes as well, since you have familiarity with InSpec.

morgante:
Replace with #34.

PR #35, opened 2018-09-29
Add migration script for G Suite refactor

adrienthebo:
This pull request adds a first pass of a migration script that can detect project-factory resources in a state file and intelligently rename them.

### DIFF
--- a/helpers/migrate.py
+++ b/helpers/migrate.py
@@ -19,6 +19,7 @@ import copy
 import subprocess
 import sys
 import shutil
+import re
 
 
 class GSuiteMigration:
@@ -153,6 +154,9 @@ class TerraformResource:
         Generate a new Terraform resource, based on the fully qualified
         Terraform resource path.
         """
+        if re.match(r'\A[\w.\[\]-]+\Z', path) is None:
+            raise ValueError("Invalid Terraform resource path {!r}".format(path))
+
         parts = path.split(".")
         name = parts.pop()
         resource_type = parts.pop()
--- a/helpers/migrate.py
+++ b/helpers/migrate.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+
+"""usage: migrate.py <old-state> <new-state>"""
+
+import argparse
+import pprint
+import subprocess
+import sys
+
+
+class TerraformModule:
+    """
+    A Terraform module with associated resources.
+    """
+
+    def __init__(self, name, resources):
+        """
+        Create a new module and associate it with a list of resources.
+        """
+        self.name = name
+        self.resources = resources
+
+    def has_resource(self, resource_type, resource_name):
+        """
+        Does this module contain a resource with the matching type and name?
+        """
+        for resource in self.resources:
+            if resource.resource_type == resource_type \
+                    and resource.name == resource_name:
+                return True
+        return False
+
+    def __repr__(self):
+        return "{}({!r}, {!r})".format(
+                self.__class__.__name__,
+                self.name,
+                [repr(resource) for resource in self.resources])
+
+
+class TerraformResource:
+    """
+    A Terraform resource, defined by the the identifier of that resource.
+
+    >>> path = 'module.project-factory.google_project.project'
+    >>> resource = TerraformResource.from_path(path)
+    >>> assert resource.module == 'module.project-factory'
+    >>> assert resource.resource_type == 'google_project'
+    >>> assert resource.name == 'project'
+    """
+
+    @classmethod
+    def from_path(cls, path):
+        """
+        Generate a new Terraform resource, based on the fully qualified
+        Terraform resource path.
+        """
+        parts = path.split(".")
+        name = parts.pop()
+        resource_type = parts.pop()
+        module = ".".join(parts)
+        return cls(module, resource_type, name)
+
+    def __init__(self, module, resource_type, name):
+        """
+        Create a new TerraformResource from a pre-parsed path.
+        """
+        self.module = module
+        self.resource_type = resource_type
+        self.name = name
+
+    def path(self):
+        """
+        Return the fully qualified resource path.
+        """
+        parts = [self.module, self.resource_type, self.name]
+        if parts[0] == '':
+            del parts[0]
+        return ".".join(parts)
+
+    def __repr__(self):
+        return "{}({!r}, {!r}, {!r})".format(
+                self.__class__.__name__,
+                self.module,
+                self.resource_type,
+                self.name)
+
+
+def group(resources):
+    """
+    Group a set of resources according to their containing module.
+    """
+
+    groups = {}
+    for resource in resources:
+        if resource.module in groups:
+            groups[resource.module].append(resource)
+        else:
+            groups[resource.module] = [resource]
+
+    return [
+            TerraformModule(name, contained)
+            for name, contained in groups.items()
+    ]
+
+
+def read_state(statefile):
+    """
+    Read the terraform state at the given path.
+    """
+    argv = ["terraform", "state", "list", "-state", statefile]
+    result = subprocess.run(argv,
+                            capture_output=True,
+                            check=True,
+                            encoding='utf-8')
+    elements = result.stdout.split("\n")
+    elements.pop()
+    return elements
+
+
+def main(argv):
+    parser = argparser()
+    args = parser.parse_args(argv[1:])
+
+    resources = [
+            TerraformResource.from_path(path)
+            for path in read_state(args.oldstate)
+    ]
+    pprint.pprint(resources)
+    modules = group(resources)
+    pprint.pprint(modules)
+    print("done")
+
+
+def argparser():
+    parser = argparse.ArgumentParser(description='Migrate Terraform state')
+    parser.add_argument('oldstate', metavar='oldstate.json',
+                        help='The current Terraform state (will not be '
+                             'modified)')
+    parser.add_argument('newstate', metavar='newstate.json',
+                        help='The path to the new state file')
+    return parser
+
+
+if __name__ == "__main__":
+    main(sys.argv)
--- a/helpers/migrate.py
+++ b/helpers/migrate.py
@@ -1,5 +1,19 @@
 #!/usr/bin/env python3
 
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


test/helpers/test_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+module.project-factory.google_service_account_iam_member.service_account_grant_to_group
+module.project-factory.null_data_source.data_final_group_email
+module.project-factory.null_data_source.data_group_email_format
+module.project-factory.null_resource.delete_default_compute_service_account
+module.project-factory.random_id.random_project_id_suffix"""
+
+
+class TestGSuiteMigration(unittest.TestCase):
+    def setUp(self):
+        self.resources = [
```
aaron-lane: `from_path` should have tests to prove its correctness, or `self.resources` should be hardcoded with the expected values so it does not depend on the correctness of `from_path`.

test/helpers/test_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+        for old in self.resources:
+            new = copy.deepcopy(old)
+            new.module += ".module.project-factory"
+            move = (old.path(), new.path())
+            assert move in moves
+
+
+class TestTerraformModule(unittest.TestCase):
+    def setUp(self):
+        self.resources = [
```
aaron-lane: Ditto regarding `from_path`.

test/helpers/test_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+            migrate.TerraformResource.from_path(path)
+            for path in TERRAFORM_STATE_LIST.split("\n")
+            if path.startswith("module.project-factory")
```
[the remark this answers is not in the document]
adrienthebo: I think that I intended to check that we didn't touch any other resources but it appears that I didn't write tests for that. Think it's worth the effort to write those tests or should we just rip out the extra resources and the `.startswith` filter?

test/helpers/test_migrate.py
adrienthebo: Added test coverage for `from_path` and some light input validation (though if `terraform state list` outputs bad values, something has gone horribly wrong)

test/helpers/test_migrate.py
```diff
@@ -0,0 +1,149 @@
...
+import unittest
+sys.path.append(
+    os.path.abspath(
+        os.path.join(
+            os.path.dirname(__file__),
+            '../../helpers/')))
+
+import migrate  # noqa: E402
+
+TERRAFORM_STATE_LIST = """google_project_iam_member.additive_sa_role
```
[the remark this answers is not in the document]
adrienthebo: Addressed in https://github.com/terraform-google-modules/terraform-google-project-factory/pull/35/commits/319f3012ba0406a1cf9e56b7c5b6545e28ece28a

test/helpers/test_migrate.py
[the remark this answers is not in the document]
adrienthebo: Addressed in https://github.com/terraform-google-modules/terraform-google-project-factory/pull/35/commits/50b5f827873b9cc33e283709b3b397c2e8143f0f

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+class GSuiteMigration:
+    """
+    Migrate the resources from a flat project factory to match the new
+    module structure created by the G Suite refactor.
+    """
+
+    CORE_PROJECT_FACTORY_SELECTORS = [
+        ("google_organization", "org"),
```
aaron-lane: `google_organization.org`  was extracted to the `google_organization` submodule as the logic around it was duplicated in `gsuite_enabled` and `core_project_factory`.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_compute_default_service_account", "default"),
+        ("null_data_source", "data_final_group_email"),
```
aaron-lane: `data_final_group_email` was refactored out of existence.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("null_data_source", "data_group_email_format"),
```
aaron-lane: `data_group_email_format` was refactored out of existence.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_compute_shared_vpc_service_project", None),
+        ("google_compute_subnetwork_iam_member", None),
```
aaron-lane: `google_compute_subnetwork_iam_member.group_role_to_vpc_subnets` remains in `gsuite_enabled` so this rule must be restricted.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_project", None),
+        ("google_project_iam_member", None),
```
aaron-lane: `google_project_iam_member.gsuite_group_role` and `google_project_iam_member.controlling_group_vpc_membership` remain in `gsuite_enabled` so this rule must be restricted.

In case we test this script against test state, `google_project_iam_member.additive_sa_role` and `google_project_iam_member.additive_shared_vpc_role` remain in `full` fixture.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_project_service", None),
+        ("google_project_usage_export_bucket", None),
+        ("google_service_account", None),
```
aaron-lane: In case we test this script on test state, `google_service_account.extra_service_account` remains in `full` fixture.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_service_account_iam_member", None),
```
aaron-lane: `google_service_account_iam_member.service_account_grant_to_group` remains in `gsuite_enabled` so this rule must be restricted.

In case we test this script on test state, `google_service_account_iam_member.additive_service_account_grant_to_group` remains in `full` fixture.

helpers/migrate.py
```diff
@@ -0,0 +1,292 @@
...
+        ("google_storage_bucket", None),
+        ("google_storage_bucket_iam_member", None),
```
aaron-lane: `google_storage_bucket_iam_member.group_storage_admin_on_project_bucket` remains in `gsuite_enabled` so this rule must be restricted.

adrienthebo:
I've refactored the migration module to better handle multiple modules, but I need to spend a lot of time handling the `gsuite_enabled` path. With the old module structure, the `gsuite_enabled` module was at the root of the module and so those resources didn't need refactoring, but the extraction of the `google_group` module broke that assumption and added new migration paths that we'll have to account for.

aaron-lane:
@adrienthebo I adjusted `aaron-lane/gsuite-refactor`; the `google_group` module wasn't actually necessary .

helpers/migrate.py
adrienthebo: I've dropped this from the refactored resource list, and since it's a data source it won't cause issues.

helpers/migrate.py
adrienthebo: I've dropped this from the refactored resource list, and since it's a data source it won't cause issues.

helpers/migrate.py
adrienthebo: The refactoring of the shared VPC `roles/compute.networkUser` role means that we can't perform a simple migration through `terraform state mv`. I'm getting this diff:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create
-/+ destroy and then create replacement

Terraform will perform the following actions:

  + module.project-factory.google_project_iam_member.controlling_group_vpc_membership
      id:      <computed>
      etag:    <computed>
      member:  "group:pf-gsuite-migrate-group@phoogle.net"
      project: "thebo-host-c15e"
      role:    "roles/compute.networkUser"

-/+ module.project-factory.module.project-factory.google_project_iam_member.controlling_group_vpc_membership[1] (new resource required)
      id:      "thebo-host-c15e/roles/compute.networkUser/group:pf-gsuite-migrate-group@phoogle.net" => <computed> (forces new resource)
      etag:    "BwV872YZKPU=" => <computed>
      member:  "group:pf-gsuite-migrate-group@phoogle.net" => "serviceAccount:303471683246@cloudservices.gserviceaccount.com" (forces new resource)
      project: "thebo-host-c15e" => "thebo-host-c15e"
      role:    "roles/compute.networkUser" => "roles/compute.networkUser"

-/+ module.project-factory.module.project-factory.google_project_iam_member.controlling_group_vpc_membership[2] (new resource required)
      id:      "thebo-host-c15e/roles/compute.networkUser/serviceAccount:303471683246@cloudservices.gserviceaccount.com" => <computed> (forces new resource)
      etag:    "BwV872YZKPU=" => <computed>
      member:  "serviceAccount:303471683246@cloudservices.gserviceaccount.com" => "serviceAccount:project-service-account@pf-gsuite-migrate-group-23dc.iam.gserviceaccount.com" (forces new resource)
      project: "thebo-host-c15e" => "thebo-host-c15e"
      role:    "roles/compute.networkUser" => "roles/compute.networkUser"
```

I think we're stuck with re-creating these resources or doing some serious state surgery, which increases the risk of something going wrong.

helpers/migrate.py
adrienthebo: These resources should only be moved when found within a project-factory Terraform module, so this won't be touched.

helpers/migrate.py
aaron-lane: Is there a risk in recreating those resources?

aaron-lane:
@adrienthebo Is there a risk in recreating those resources?

adrienthebo:
I've rebased this on top of `aaron-lane/gsuite-refactor`, so a reset against the updated branch may be required. In addition I've verified that the recreation of the IAM roles does not create an outage; the most I saw was a very brief increase in the instance ping time but no real interruption.

adrienthebo:
I've updated the tests and used the output of `terraform state list` from a test environment as fixtures. The tests feel a little contrived though, could you eyeball them and see if we're actually testing anything interesting when checking the computed moves?

aaron-lane:
@adrienthebo: I don't see any major issues with the tests. The stubbed state list of resources could be shortened for clarity; I don't think 17 counts of the same resources really provide any more value than say 2 counts.

adrienthebo:
@aaron-lane fair point - I've trimmed that down to 3 subnets, and added some additional tests to make sure that we don't touch resources in other modules. Care to give another look?

PR #36, opened 2018-10-02
Add script for generating root level module

adrienthebo:
This commit adds `helpers/generate_root_module/generate_root_module.py` that generates a top level Terraform module, based on the variables and outputs of `modules/core_project_factory`. It can be invoked via `make generate` or by `./helpers/generate_root_module/generate_root_module.py`.

This commit has been rebased down to a single unit of work and contains a number of improvements and fixups by Aaron Lane <aaronlane@google.com>

- [ ] Install pip dependencies in unit test docker image


aaron-lane:
This must be rebased against master once #73 is resolved.

adrienthebo:
Rebase done.

helpers/generate_root_module.py
```diff
@@ -0,0 +1,194 @@
+#!/usr/bin/env python3
+
+import re
+import subprocess
+import sys
+
+
+OUTPUT_OBJECT_PATTERN = re.compile(r"""
```
adrienthebo: @aaron-lane one of the things I'd like reviewed or discussed is this regular expression. Since I'm trying to parse a context free grammar with a regular expression this is quite brittle, and I think we'd be better served by using a Python HCL parser to handle this - but that'll mean pulling in an outside dependency which is also problematic. What's your take on this approach?

aaron-lane: I think using a package like [pyhcl](https://github.com/virtuald/pyhcl) is preferable to a complex regular expression. pyhcl in particular only depends on [ply](https://github.com/dabeaz/ply/tree/master) which appears to have no dependencies of its own. Does that satisfy the concern about external dependencies?

adrienthebo: Yeah, anything is better than this regular expression. We can throw a `requirements.txt` in either `.` or `./helpers` to handle the dependency definition, do you have a preference?

aaron-lane: I think we should add a second directory under `./helpers` to isolate this particular script and its requirements from other scripts.

morgante:
@adrienthebo Could you fix conflicts with master?

aaron-lane:
The root module is no longer simple enough to generate with this approach.

PR #41, opened 2018-10-04
Add user control over deleting default compute user

adrienthebo:
Add user control over deleting default compute user

A number of GCP modules such as the [terraform-google-kubernetes-engine][gke]
module default to using the default compute user account, and the
deletion of the default compute user account makes using the
project-factory more complicated. A user's first experience with either
the project-factory or another Terraform that uses GCE resources will
have the following experience:

- User finds an external Terraform module to solve a problem - eg, gke.
- User finds an example usage of that module that requires a `google_project`
- User creates a project-factory based on a minimal example from this module
- Terraform creates a new `google_project`
- Terraform deletes the default compute service account
- User creates a gke module based on a minimal example from that module
- Terraform tries to creates gke resources with a default compute user dependency
- The gke resources fail to create because of the missing dependency
- Terraform crashes out with a confusing error

While we can reasonably opt for a security-first approach we can make
this deletion the default behavior, but we should provide an escape
hatch. Allowing the compute user to be kept means that it'll be a lot
easier for users to spin up prototypes.

[gke]: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -132,6 +132,7 @@ data "google_compute_default_service_account" "default" {
   Default compute service account deletion
  *****************************************/
 resource "null_resource" "delete_default_compute_service_account" {
+  count = "${var.delete_default_compute_user}"
   provisioner "local-exec" {
     command = "${path.module}/scripts/delete-service-account.sh ${local.project_id} ${var.credentials_path} ${data.google_compute_default_service_account.default.id}"
   }
--- a/variables.tf
+++ b/variables.tf
@@ -129,3 +129,8 @@ variable "app_engine" {
   type        = "map"
   default     = {}
 }
+
+variable "delete_default_compute_user" {
+  description = "Remove the default compute service account"
+  default     = true
+}


adrienthebo:
@ryanckoch pinging you for review on this, have you had this experience in your work as well?

morgante:
I would prefer to avoid this. Personally, I think the dependency on the default Service Account is insecure and suboptimal.

Instead, we should be ensuring all our modules do *not* have that dependency. @ryanckoch I thought we had worked to make sure the GKE module didn't need a default Service Account?

My strong preference is we focus work on the other modules rather than re-introducing the default Service Account.

adrienthebo:
@morgante I think this conversation is worth having, because this is forcing a pretty painful opinion on end users.

A good example of this came up when I tried to use the project-factory to create a project for testing inspec-gcp. Inspec-gcp uses Terraform and https://github.com/GoogleCloudPlatform/terraform-google-managed-instance-group to stand up test fixtures. Inspec-gcp uses the managed-instance-group module in the way that most of our users would - by using the default values that stand up compute resources which in turn depend on the deleted compute user. This issue is only encountered after a ~10 minute Terraform run and usually leaves the resulting project in a mangled state.

Now, granted, the managed-instance-group module can use a specific service account - but doing so requires modifying inspec-gcp's tests. This is the workflow that users repeat constantly - rely on the project-factory, have something break, spend time chasing down the issue, and then decide whether to patch in a non-default user.

I submitted this change because this is more than a small frustration - it significantly degrades the value of the project-factory with nearly every external module I've used that interacts with the compute or container APIs, and one constantly runs into this problem.

If users run into this pain point frequently enough (and I believe that they will), then they'll give up and create their own project-factory. This scenario is the worst of both worlds - users aren't using the project-factory which represents a lot of sunk cost on a module nobody's using, and home-grown project factories are much more likely to preserve the default compute user.

All in all, by increasing the barrier to entry to the project-factory, I think that we'll push people to use less secure options by default. This is why `curl -k` was so prevalent for a while - getting TLS certificate validation was sufficiently painful that people turned off authentication entirely and exposed themselves to MITMs.

Alternately I might be over-estimating how much pain this may cause end users - I've been wrong about things like this before. But it would be nice to get a feel for how this may positively (or negatively) affect users.

morgante:
@adrienthebo I'm happy to discuss this more.

The issue is that the default Service Account is the equivalent of `curl -k`. It's easy to get started with but exposes you to a whole host of security problems. Out of the box, we should be encouraging best practices (and all of our modules should reflect this).

We should be encouraging this best practice in the ecosystem and not encouraging insecure defaults.

One alternative which I would like to see us pushing across the board is to integrate *creating* a Service Account when none is provided instead of reusing the default account. We could integrate that into the GKE and MIG modules.

> spend time chasing down the issue, and then decide whether to patch in a non-default user.

This seems very bad - if you try to do something which depends on the default user and it doesn't exist, we should fail fast and give an immediate clear error.

Before considering introducing an unsafe feature, I'd like to make sure we've done all we can to make sure this is unnecessary:

- Making sure all our modules do not require a default Service Account (can you confirm this for GKE module)
- Providing easy guidance on not using a Default Service Account
- Including documentation on the anti-patterns associated with it

adrienthebo:
> We should be encouraging this best practice in the ecosystem and not encouraging insecure defaults.

Yes, definitely, and the fact that I'm arguing against the more secure option is probably a good argument *against* this pull request.

> This seems very bad - if you try to do something which depends on the default user and it doesn't exist, we should fail fast and give an immediate clear error.

One example of where the default compute user is used by default is in the [`google_compute_instance`](https://www.terraform.io/docs/providers/google/r/compute_instance.html#email) resource. The [`google_container_cluster`](https://www.terraform.io/docs/providers/google/r/container_cluster.html#service_account) resource depends on it by default as well. I suspect that there are more, but by default if you use these resources with a minimal set of options with a project created by project-factory, these resources will fail.

Now of course, per your argument that doesn't mean that we should restore the default compute user, but it does mean that the workflow isn't great.

I think the biggest pain point isn't the existence of the user; it's how long it takes for things to fail. [tflint](https://github.com/wata727/tflint) exists and might be a great way to solve this problem - but it doesn't support GCP resources.

After thinking about it, I think that providing a linter to catch issues like this would be a much better solution in the long run, albeit more labor intensive in the short run.

What are your opinions on closing this pull request in favor of putting work into a linter for Terraform GCP resources?

morgante:
> I'm separately interested in a Terraform linter, but I think we'd need a clearer scope of 

Totally agreed, and this is what I'm curious about. If it's taking a long time to fail, we should file upstream bugs on the provider (and our own modules). As immediate next steps I would suggest:

- PR for `inspec-gcp` to fix the dependency on default SA
- issue for Terraform provider to not require and/or fail fast on non-existent Service Account for `google_compute_instance` and `google_container_cluster`.
- Issue for our GKE module to remove all dependencies on default Service Account; none of our modules should use the default one at all

Separately, I'm interested in working on a linter of a some sort and it will be a feature we likely eventually add to the Cloud Foundation Toolkit.

Issue #29, opened 2018-09-18
detach shared vpc service project from host shared vpc prior to deletion


reechar-goog:
https://cloud.google.com/vpc/docs/deprovisioning-shared-vpc#deleting_shared_vpc_service_project

When deleting a GCP project that is a service project associated with a Host Shared VPC project, terraform should attempt to remove the association before deleting the project.

Jberlinsky:
@reechar-goog It looks to me like this is a non-issue -- although I'm curious if you have a sample case that reproduces a specific issue?

When a `terraform destroy` is run in this repository, among other things, the following will execute in order (as enforced by `depends_on` statements:

1. The `google_compute_shared_vpc_service_project` resource will be deleted, invoking [https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_shared_vpc_service_project.go#L116](a detachment of the service project from the shared VPC)
2. The `google_project_service` resources will be deleted, and finally
3. The `google_project` resource will be deleted.

If you have a case where this doesn't match your expectations, I'd love to hear about it and see what I can do to help! Otherwise, I'm going to close this issue out. Please feel free to re-open if you have more to add.

PR #43, opened 2018-10-04
Detach shared vpc before deleting

Jberlinsky:
Addresses https://github.com/terraform-google-modules/terraform-google-project-factory/issues/29

Note that:

1. This is to be merged into `morgante/kitchen-docker` per conversation with @adrienthebo 
2. Until https://github.com/terraform-google-modules/terraform-google-project-factory/pull/42 is merged in and rebased, it will _appear_ that this change breaks project destruction.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -119,6 +119,11 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
   service_project = "${local.project_id}"
 
   depends_on = ["google_project_service.project_services"]
+
+  provisioner "local-exec" {
+    when = "destroy"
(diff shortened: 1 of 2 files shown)


morgante:
Do we require a bash script to do this? I think we need to do more discovery on exactly what `gcloud compute shared-vpc associated-projects remove` does and how it differs from what Terraform does on the "destroy" operation for the `shared_vpc_attachment` resource.

Additionally, we need to be sure this has automated testing to identify the necessity of this script and detect and problems with it.

Jberlinsky:
Follow-up here: https://github.com/terraform-google-modules/terraform-google-project-factory/issues/29#issuecomment-427124542

PR #44, opened 2018-10-05
Fixed bugs of the centos helper install script.

garzoglio:
- dep installation did not sudo the executing shell; env var were not
passed.
- terraform-provider-google required make fmt before sucessful build.
Installation also failed (bug in cp command).
- Some clean up


garzoglio:
Fixed one additional bug that installed the terraform gsuite provider under ~root/.terraform.d instead of ~/.terraform.d
Also changed the name of the test configuration file from sample.sh to config.sh, as expected by the test scripts

morgante:
@garzoglio Thanks for the contribution - do you mind running `make -s` and fixing any linting errors in the centos script while you're working on it?

garzoglio:
Fixed linting errors and tested that the script installs the environment properly

PR #46, opened 2018-10-08
Test project-factory with different sets of variables

adrienthebo:
This pull request is based on #34.

This pull request restructures the test-kitchen tests implemented in #34 to define two platforms - a "full" platform and a "minimal" platform. These separate configurations allow us to test the project-factory in two manners - one with as many variables set as provided by the user, and one with the bare minimum number of variables set.

The intent of this refactor is to ensure that we have adequate test coverage of the project-factory features, while ensuring that variables with defaults are truly optional.

The use of both platforms is not required; if a feature is under development that requires a full set of variables, test-kitchen can be restricted to a single platform:

```sh
kitchen create default-full
kitchen converge default-full
kitchen converge default-full
kitchen verify default-full
kitchen destroy default-full
```

The downside of this approach is that test-kitchen will default to standing up two projects in order to  verify all code paths, and projects are not an infinite resource.


adrienthebo:
I've verified this change by reverting 5af6f59436a5a199380d3389b1f7da1c54bc7852 and then attempting to create the `default-minimal` platform with test-kitchen. Test-kitchen fails with the reverted commit, and passes with the fix applied.

adrienthebo:
Closing in favor of #34.

PR #48, opened 2018-10-10
Add troubleshooting guide

adrienthebo:
[Rendered markdown](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/adrienthebo/troubleshooting-guide/docs/TROUBLESHOOTING.md)

The Project Factory interacts with a number of APIs and requires fairly
broad permissions and a number of enabled APIs in order to operate. If
some of these permissions or services or absent the Project Factory can
fail to run, or partially generate a project and then get stuck.

This commit adds documentation for some of the more common errors and
misconfigurations, explains the cause of the errors, and indicates how
to fix them.

This document has gone through a few iterations, so the wording may need some work. There are some cases where I expected an error, such as with a missing `compute.googleapis.com` API on the seed project, but things didn't fail. This might mean that the dependency doesn't exist, or my testing was faulty. These may warrant closer review.


TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
+# Troubleshooting
```
morgante: Let's put this at docs/TROUBLESHOOTING.md.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+* **seed project**: An existing GCP project with resources, services, and service accounts needed to create projects with the project factory.
+* **service account**: The service account (which exists in the **seed project**) that creates the **target project**.
+* **target project**: The project being generated by the Project Factory.
+
+- - -
+## Problems
+
+* [Seed project missing APIs](#seed-project-missing-apis) - The seed project is missing required APIs.
+* [Service account missing roles](#service-account-missing-roles) - The service account has insufficient permissions.
+* [Common issues](#common-issues)
```
morgante: Should we put Common Issues at the top of the list?

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+
+### Seed project missing APIs
+
+The Project Factory requires the following services are enabled on the seed project. If these APIs are not enabled they can cause the Project Factory to get into a bad state.
```
morgante: Let's include a link to the README section which provides the exhaustive list, in case this gets out of sync.

https://github.com/terraform-google-modules/terraform-google-project-factory/tree/adrienthebo/troubleshooting-guide#apis

Also let's add links from the README back to this FAQ for each of the APIs (I liked that you included an explanation of why the API is needed.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+$ gcloud services enable cloudbilling.googleapis.com --project "$SEED_PROJECT"
+```
+
+**Notes:**
+
+This error will occur *once* when applying a Terraform plan when the **seed project** does not have the Cloud Billing API enabled. On subsequent Terraform runs, Terraform will generate an error indicating that the Compute Engine API is not enabled on the **target project**. Watch out for this!
+
+### Missing API: `compute.googleapis.com`
+
+TODO: it appears that the `compute` API isn't actually used by the project factory. Can we confirm this?
```
morgante: I believe we use it to destroy the default network? I'd prefer to not remove it from the list but we can remove the Troubleshooting section.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+
+* module.project-factory.google_service_account.default_service_account: 1 error(s) occurred:
+
+* google_service_account.default_service_account: Error creating service account: googleapi: Error 403: Identity and Access Management (IAM) API has not been used in project [seed-project-number] before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/iam.googleapis.com/overview?project=[seed-project-number] then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured
+```
+
+**Cause:**
+
+The seed project does not have the `iam.googleapis.com` API enabled. This prevents the Project Factory from deleting the default GCE service account.
```
morgante: Also prevents it from granting access to groups, etc.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+
+```
+* module.project-factory.data.google_organization.org: 1 error(s) occurred:
+
+* module.project-factory.data.google_organization.org: data.google_organization.org: Error reading Organization Not Found : [organization-id]: googleapi: Error 403: The caller does not have permission, forbidden
+```
+
+**Cause:**
+
+The active service account does not have `roles/resourcemanager.organizationViewer` on the active organization.
```
morgante: We should also include an explanation of why each role is required, and link to this explanation from the README section.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,420 @@
...
+# Requires `roles/serviceusage.serviceUsageAdmin` on $SEED_PROJECT
+gcloud services enable appengine.googleapis.com --project "$SEED_PROJECT"
+```
+
+**Notes:**
+
+The absence of the `appengine.googleapis.com` API will not cause Terraform to fail, but any interactions with the App Engine instance will fail. Once this API is activated, it might take a few minutes for Terraform to detect that the App Engine instance is not present and then create it.
+
+- - -
+### Service account missing roles
```
morgante: We should include a high-level link back to the [README](https://github.com/terraform-google-modules/terraform-google-project-factory/tree/adrienthebo/troubleshooting-guide#permissions).

TROUBLESHOOTING.md
adrienthebo: I was surprised to find that the README doesn't state `compute.googleapis.com` as an API - I always assumed it was there. Will delete this section from the troubleshooting guide.

PR #53, opened 2018-10-18
Implement Billing Account Role

knitterb:
This adds to the helper script the optional creation of the IAM billing role for the service account that is required.  Other small changes include:

- Additional output when the hep script is run so that the user understands what steps are being executed.  This could be helpful for any debugging or failure.
- Explicit details on how to add a service account to the gcloud auth list

### DIFF
--- a/helpers/setup-sa.sh
+++ b/helpers/setup-sa.sh
@@ -18,9 +18,13 @@ set -e
 set -u
 
 # check for input variables
-if [ $# -ne 2 ]; then
+if [ $# -lt 2 ]; then
   echo
-  echo "Usage: $0 <organization name> <project id>"
+  echo "Usage: $0 <organization name> <project id> [<billing account id>]"
+  echo
+  echo "  organization name (required)"
+  echo "  project id (required)"
+  echo "  billing account id (optional)"
   echo
   exit 1
 fi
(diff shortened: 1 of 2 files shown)


helpers/setup-sa.sh
```diff
@@ -130,4 +157,28 @@ gcloud services enable \
...
+    sed -i '' -e "/^etag:.*/i \\
+- members:\\
+\ \ - serviceAccount:${SA_ID}\\
+\ \ role: roles/billing.user" policy-tmp-$$.yml
+    gcloud beta billing accounts set-iam-policy $BILLING_ACCOUNT policy-tmp-$$.yml
+  else
+      echo "Could not set roles/billing.user on service account $SERVICE_ACCOUNT.\
+      Please perform this manually."
+  fi
+  rm -f policy-tmp-$$.json
```
adrienthebo: The extension of this file is incorrect, it should be `.yml` but is `.json`.

helpers/setup-sa.sh
```diff
@@ -130,4 +157,28 @@ gcloud services enable \
...
+\ \ role: roles/billing.user" policy-tmp-$$.yml
+    gcloud beta billing accounts set-iam-policy $BILLING_ACCOUNT policy-tmp-$$.yml
+  elif [ "$unamestr" = 'Linux' ]; then
+    sed -i '' -e "/^etag:.*/i \\
+- members:\\
+\ \ - serviceAccount:${SA_ID}\\
+\ \ role: roles/billing.user" policy-tmp-$$.yml
+    gcloud beta billing accounts set-iam-policy $BILLING_ACCOUNT policy-tmp-$$.yml
+  else
+      echo "Could not set roles/billing.user on service account $SERVICE_ACCOUNT.\
```
adrienthebo: Small fixup: since we used a 2 space indent in the previous code, could we switch this block to use 2 spaces as well?

knitterb:
Fixed both requested change @adrienthebo

Issue #27, opened 2018-09-17
Terraform state deadlock with a limited host project


adrienthebo:
## Summary

The project-factory can use a service account with a misconfigured host project, which leads to a deadlocked Terraform configuration with confusing errors.

## Description

The project factory requires that the host project associated with an instantiating service account has certain APIs enabled, but the project factory does not verify that these APIs are enabled before generating a new project. This can cause the project factory to partially generate a new project and then fail, and subsequent calls to `terraform apply` will fail with a number of related but unhelpful errors.

When `terraform apply` is run with the misconfigured environment, it will create a new project, attempt to associate a billing account with that project, and fail since `cloudbilling.googleapis.com` is disabled. Following runs of `terraform apply` will see that the google_project was created and then attempt to look up the default compute service account with the `google_service_account` data source. This will fail because the `compute.googleapis.com` API is not enabled, but this hides the more interesting errors - that the service project doesn't have a billing account enabled, and that the host project doesn't have the Cloud Billing API enabled and thus can't remedy the issue through Terraform.

## Expected behavior

When the project-factory runs with a service account whose host project doesn't have the right APIs enabled, it terminates before generating resources.

## Actual behavior

First invocation of `terraform apply`:

```
module.project-factory.google_project.project: Still creating... (10s elapsed)

Error: Error applying plan:

1 error(s) occurred:

* module.project-factory.google_project.project: 1 error(s) occurred:

* google_project.project: Error setting billing account "<BILLING ACCOUNT>" for project "projects/tftest-1-4901": googleapi: Error 403: Cloud Billing API has not been used in project <HOST NUMBER> before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudbilling.googleapis.com/overview?project=<HOST NUMBER> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry., accessNotConfigured

Terraform does not automatically rollback in the face of errors.
Instead, your Terraform state file has been partially updated with
any resources that successfully completed. Please address the error
above and apply again to incrementally change your infrastructure.
```

**Note:** that the error generated references the **host project** number on the first failed run.

Running `terraform plan` or `terraform apply`
```
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.

random_id.random_project_id_suffix: Refreshing state... (ID: SQE)
data.google_organization.org: Refreshing state...
[...]

PR #54, opened 2018-10-18
Add project-factory preconditions checker script (#27)

adrienthebo:
The Project Factory requires a number of enabled APIs and granted
permissions on the service account. The absence of these requirements
can cause the Project Factory to fail to run, generate invalid projects,
and fail in difficult to debug ways.

This pull request adds a python script to validate that the Project Factory
preconditions are met, based on the configuration given to the script.

This pull request does not yet integrate this script into the Project Factory
itself, because it has python dependencies that may not be satisfied when the
Project Factory is pulled in via Terraform's dependency manager. The intent of
this pull request is to validate this approach to validating the Project
Factory preconditions; if this approach is accepted we can look into further
integrations and automation.


adrienthebo:
@morgante it appears that `gcloud` does not implement the `testIamPermissions` on any resources. Without this implementation I don't think there's a good way of verifying permissions without querying the IAM roles, which in turn will require that the service account has the `getIamPolicy` permission on a number of resources.

How would you like to proceed with this feature?

morgante:
@adrienthebo That is unfortunate.

Could we make this optional? ie. have it wrapped in something which will check if you have the requisite dependencies and only run if you do?

adrienthebo:
That's the direction where I was leaning. My thinking was that we could have a helper script that would probe for python and the required packages, run the verifier if possible, warn/exit otherwise.

I'll put together a proof of concept and update this pull request.

adrienthebo:
I've added a Bourne wrapper script that will check for `python` before running the preconditions checker, so this pull request adds no new dependencies.

morgante:
@adrienthebo What is Bourne?

adrienthebo:
Bourne -> `sh` as opposed to `bash`, since we can't rely on environments such as Alpine Linux providing bash. This script should run on both 2.7 and 3.7.

aaron-lane:
@morgante I successfully executed the tests with Python 2 and Python 3.

scripts/preconditions/preconditions.py
```diff
@@ -0,0 +1,413 @@
...
+            "Service account permissions on parent folder",
+            resource,
+            self.permissions,
+            response["permissions"],
+        )
+
+        return req.asdict()
+
+
+class HostVpcProjectPermissions:
```
aaron-lane: Should this be renamed to `SharedVpcProjectPermissions` to match the intended usage?

scripts/preconditions/preconditions.py
```diff
@@ -0,0 +1,421 @@
...
+    parser.add_argument(
+        '--billing_account', required=True,
+        help='The billing account to be associated with a new project'
+    )
+    parser.add_argument(
+        '--org_id', required=True, action=EmptyStrAction,
+        help='The organization ID'
+    )
+    parser.add_argument(
+        '--folder_id', required=False, action=EmptyStrAction,
```
aaron-lane: `folder_id` is not used to verify folder permissions.

scripts/preconditions/preconditions.py
adrienthebo: Sure, would you like me to make that change or would you prefer to do it?

main.tf
```diff
@@ -83,6 +83,31 @@ data "google_organization" "org" {
   organization = "${var.org_id}"
 }
 
+resource "null_resource" "check_preconditions" {
+  triggers {
+    credentials_path = "${var.credentials_path}"
```
adrienthebo: We should run `terraform fmt` to fix up this formatting.

main.tf
```diff
@@ -83,6 +83,31 @@ data "google_organization" "org" {
...
+${path.module}/scripts/check-preconditions.sh \
+    --credentials_path '${var.credentials_path}' \
+    --billing_account '${var.billing_account}' \
+    --org_id '${var.org_id}' \
+    --folder_id '${var.folder_id}' \
+    --shared_vpc '${var.shared_vpc}'
+EOD
+
+    environment {
+      GRACEFUL_IMPORTERROR = true
```
adrienthebo: For consistency's sake we should quote true.

main.tf
```diff
@@ -83,6 +83,31 @@ data "google_organization" "org" {
   organization = "${var.org_id}"
 }
 
+resource "null_resource" "check_preconditions" {
```
adrienthebo: Since we renamed the preconditions script I think we should rename this to be consistent.

scripts/preconditions/preconditions.py
adrienthebo: It's been months since I've touched this pull request so my memory is foggy, but I recall running into issues where testing for the `resourcemanager.folders.create` permission required walking the parent folders and testing for the permission. I'm happy to revisit this, or remove the check entirely - what's your thinking?

aaron-lane: I added the validation back in to the logic and it appears to work. I tested it against an invalid folder ID and a valid folder ID with and without the `resourcemanager.folders.get` permission.

PR #56, opened 2018-10-24
added google style project id generator

ferrier1:
#!/bin/bash -e
MYDIR="$(dirname "$(readlink -f "$0")")"
word1="$(shuf -n1 "${MYDIR}"/dictionary.txt)"
word2="$(shuf -n1 "${MYDIR}"/dictionary.txt)"

jq -n --arg word1 "${word1}" --arg word2 "${word2}" '{"word1":$word1,"word2":$word2}'

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,6 +128,12 @@ In order to operate the Project Factory, you must activate the following APIs on
 - Admin SDK - `admin.googleapis.com`
 - Google App Engine Admin API - `appengine.googleapis.com`
 
+## Caveats
+
+### Moving projects from org into a folder
+
+There is currently a bug with moving a project which was originally created at the root of the organization into a folder. The bug and workaround is described [here](https://github.com/terraform-providers/terraform-provider-google/issues/1701), but as a general best practice it is easier to create all projects within folders to start. Moving projects between different folders *is* supported.
+
 ## G Suite
 The Project Factory module *optionally* includes functionality to manage G Suite groups as part of the project set up process. This functionality can be used to create groups to hold the project owners and place all Service Accounts into groups automatically for easier IAM management. **This functionality is optional and can easily be disabled by deleting the `gsuite_override.tf` file**.
 


morgante:
@ferrier1 Can you clarify the intent of this PR?

ferrier1:
Google has started using 2 random words totally unrelated to the project name for the project id:

project_name: project a
project_id: random-word-91237

This PR creates project_id's that follow this schema.

It uses an external data source (bash script) that uses a dictionary and returns json items using jq.

morgante:
@ferrier1 It looks like your change doesn't include any actual code though.

In terms of the underlying change, I'm not sure if completely random project IDs is a best practice. I'd suggest that instead we only do that in cases where the `random_project_id` input is `true`.

morgante:
Also, instead of using a bash script and the complexities which come with that, I'd suggest simply using the random_pet resource to get two random pet names: https://www.terraform.io/docs/providers/random/r/pet.html

ferrier1:
Good suggestion, I have added these in and removed the bash script. Thanks

morgante:
Looks like this has merge conflicts, could you fix those?

main.tf
```diff
@@ -17,40 +17,49 @@
...
+
+ /******************************************
+   Locals configuration
+  *****************************************/
+ locals {
+   project_id             = "${google_project.project.project_id}"
+   project_number         = "${google_project.project.number}"
+   project_org_id         = "${var.folder_id != "" ? "" : var.org_id}"
+   project_folder_id      = "${var.folder_id != "" ? var.folder_id : ""}"
+   temp_project_id        = "${var.random_project_id ? format("%s-%s", random_pet.name.id, random_integer.suffix.id) : var.name}"
```
adrienthebo: This makes `var.name` optional by supplying a fallback value, but doesn't make the `name` variable itself optional. Is this intentional?

main.tf
```diff
@@ -17,40 +17,49 @@
...
-  app_engine_config = {
-    enabled  = "${list(var.app_engine)}"
-    disabled = "${list()}"
-  }
-}
+ resource "random_pet" "name" {
+
+ }
+
+ resource "random_integer" "suffix" {
```
adrienthebo: Out of curiosity, why was this switched from `random_id` to `random_integer`? Both should work equally well, but I'd like to see if there's a particular reason for this change.

ocervell:
Doesn't this PR remove the ability to have project name specified and a random suffix appended ?

In the current version, I can specify a project name (myorg-example) and set up a random suffix to facilitate destroy / recreate. 

It seems like with this PR we would only be able to autogenerate the complete name with the suffix, but doesn't support the use case I just described.

main.tf
ocervell: +1, also see my comment on this

adrienthebo:
@ocervell Good point about the inability to set a specific project ID! I sketched out this truth table, does this cover the range of behaviors we want to allow?

| `var.name` | `var.random_project_id` | `local.temp_project_id` |
| --- | --- | --- |
| `"some-project-id"` | `true` | `"some-project-id-${random-suffix}"` |
| `""` | `true` | `"${random-pet}-${random-suffix}"` |
| `"some-project-id"` | `false` | `"some-project-id"` |
| `""` | `false` | Terraform error |

main.tf
ferrier1: I wanted to create the project ID to be as close to a google generated one. They seem to be using a 6 digit integer. But yes both work equally well.

main.tf
ferrier1: var.name is now optional and uses the pet-name without the random int on the end.
null_resource validation has also been added.

ferrier1:
I cant see the reason behind setting the random_id value to true and not getting a random ID. Unless its important to have the project name and project ID identical?

Below is the current truth table after the first round of changes.


var.name | var.random_project_id | local.temp_project_id
-- | -- | --
"some-project-id" | true | "${random-pet}-${random-suffix}"
"" | true | "${random-pet}-${random-suffix}"
"some-project-id" | false | "some-project-id"-${random-suffix}"
"" | false | Terraform error

morgante:
@ferrier1 I'd like to understand your use case better before we merge something like this.

In general, we find it is much easier to manage projects if the project ID is close or identical to the project name.

That also mirrors the behavior in the Cloud Console. In the Cloud Console, entering the following names would yield these project IDs:


project name | project ID
-- | --
my-very-unique-name | my-very-unique-name
my-cool-project | my-cool-project-221314
my-project | tensile-ship-221314

As is, the only scenario the project factory doesn't reflect is the final one (where a very generic project name is used and hence the entire project ID is random).

Is that a major pain point for you? I'm not sure if the additional complexity of supporting scenario 3 is worth the additional code.

adrienthebo:
@ferrier1 I wanted to check in on this pull request - are you still interested in working on this pull request? Could you clarify the use cases you've encountered that this pull request resolves?

PR #57, opened 2018-10-25
Start a FAQ

morgante:


garzoglio:
This helps clarify some of the questions, thank you.

I suggest to add maybe one sentence next to "Delete the default compute service account" to explain why it is best practice, since then we recreate one.

The section on "Role granted" explains the membership of the API SA to api_sa_group: that's good.

morgante:
@garzoglio Thanks, I added a note about the default Service Account here: b5e15ca1545e336b7c991305bc5a68e1db7836a7

PR #67, opened 2018-11-15
Fixup errors in test-kitchen test coverage

adrienthebo:
When the test-kitchen tests and fixture modules were refactored and different fixtures were broken out for minimal and full suites, some tests were broken but were hidden because those tests were optional - that is, the test groups were disabled when the relevant parameters were unset.

In addition it was assumed that `inspec` would catch some of these errors by requiring certain inspec attributes would be set, but it appears that the version of inspec doesn't error out when required attributes were missing.

This pull request corrects a number of mistakes from string keys used instead of symbols, and passing through variables into outputs so that inspec can access them.

An outstanding question with this pull request is how to test features that require setup that are particularly heavyweight and/or can't be handled by Terraform. The original approach was to make these features optional, and if the module under test had the appropriate variables defined, then we'd test them. However the test failures that this pull request fixes shows that this approach may not be viable in the long term.

We have the following dependencies that are are optional due to non-trivial setup:

- A shared VPC host project
  - Do we want to stand up a new project with a VPC every time we test the `full` suite? Projects are a somewhat limited resource, but not too limited.
- G Suite group management
  - Adding a group to IAM roles is easy, but managing G Suite group membership cannot be done by Terraform - a user has to manually configure the service account running the tests with domain wide delegation.
  - Adding a group to/from IAM roles is easy, but the tests need a test fixture group to work with. We can create a group on the fly, but that'll require domain-wide delegation
- A destination GCS bucket for GCP project usage
  - Creating a bucket is cheap, but the bucket needs to be created within a project. Should we just stand up a fixture project for this and the shared VPC host project?

This pull request doesn't try to solve these problems; if a dependency is missing we'll just skip the respective tests - but I think we'll need to solve this for other modules.


test/integration/full/controls/gsuite.rb
```diff
@@ -40,14 +40,8 @@
...
-    it "includes the gsuite group in the given role" do
-      if !group_role
-        pending
-      end
-
-      binding = bindings.find { |b| b['role'] == group_role }
-      expect(binding).to_not be_nil
-      expect(binding[:members]).to include("group:#{group_email}")
+    it "includes the group #{group_email.inspect} in role #{group_role.inspect}" do
+      expect(bindings).to include members: including("group:#{group_email}"), role: group_role
```
morgante: Nit. Is it possible to split this onto multiple lines for readability?

test/integration/full/controls/gsuite.rb
```diff
@@ -63,16 +57,13 @@
       end
     end
 
-    it "includes the group email " do
-      binding = bindings.find { |b| b['role'] == 'roles/iam.serviceAccountUser' }
-      expect(binding).to_not be_nil
-      expect(binding[:members]).to include "group:#{group_email}"
+    it "includes gsuite group #{group_email.inspect} in the binding 'roles/iam.serviceAccountUser'" do
+      expect(bindings).to include members: including("group:#{group_email}"), role: 'roles/iam.serviceAccountUser'
```
morgante: Nit, separate lines for readability.

test/integration/full/controls/shared-vpc.rb
```diff
@@ -48,36 +48,46 @@
     end
 
     describe "roles/compute.networkUser" do
-      let(:binding) do
-        bindings.find { |b| b['role'] == 'roles/compute.networkUser' }
-      end
-
       it "includes the project service account in the roles/compute.networkUser IAM binding" do
-        expect(binding[:members]).to include "serviceAccount:#{service_account_email}"
+        expect(bindings).to include(
```
morgante: This format is good, easier to read. Please use everywhere.

PR #72, opened 2018-11-22
Invert comparison when managing project network membership

adrienthebo:
An inverted comparison caused us to not add members to the
`roles/compute.networkUser` role when shared VPC subnets were supplied.
This commit fixes the error.

Thanks to @aaron-lane for improving the test-kitchen test coverage to catch this!


aaron-lane:
I verified this by locally cherry picking it on to #67.

adrienthebo:
Thank you!

PR #103, opened 2019-01-11
Small improvements to migration docs.

morgante:


docs/upgrading_to_project_factory_v1.0.md
```diff
@@ -260,6 +264,20 @@ After restoring remote state, you need to re-initialize Terraform and push your
...
+Once you are done with the migration, you can safely remove `migrate.py`.
+
+```
+rm migrate.py
+```
+
+If you are using remote state, you can also remove the local state copies.
+
+```
+rm -rf terraform.tfstate*
```
adrienthebo: Small detail - this might be safer if we suggest `terraform.tfstate.*`, since that pattern won't inadvertently blow away `terraform.tfstate`. Your call on if we should make this change.

morgante: If you're using remote state, you don't want `terraform.tfstate` laying around anyways.

PR #118, opened 2019-01-17
Slight doc tweaks

morgante:


morgante:
@aaron-lane No, just removing the references from the README.

PR #119, opened 2019-01-17
Make Targets to Run Preconditions and Migrate Script Tests

trotttrotttrott:
These are newly added tests. Make targets are added to run them and those targets have been added as dependencies to the `all` target. Being dependencies of `all` will make them run as part of the `lint-tests` job of the project-factory Concourse CI pipeline.

Something to note about these tests is that upon failure, they will cause Make to exit as opposed to proceeding with running subsequent target dependencies. This is different than most of the others, but I would argue its behavior is preferable as it will cause tests to fail faster.

As a followup, I intend to update the project-factory Concourse pipeline so that it installs the necessary Python package dependencies before test execution.


trotttrotttrott:
Closing for the time being to refactor in response to a comment from @morgante in separate communication channel:

> Getting this into CI is a good idea, but it should be separate from the linting tests. We could add a new pipeline for unit tests or somehow incorporate, but unit tests shouldn't be entangled with linting.

trotttrotttrott:
A new PR was created for refactor: https://github.com/terraform-google-modules/terraform-google-project-factory/pull/120.

Reopening was not possible due to a force-push I had done in the process. Apologies for the noise.

PR #128, opened 2019-01-21
Remove generate_docs from default make target

aaron-lane:
Generating documentation is unnecessary during continuous integration
builds.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ DOCKER_REPO_BASE_KITCHEN_TERRAFORM := ${DOCKER_ORG}/cft/kitchen-terraform:${DOCK
 
 all: check_shell check_python check_golang check_terraform check_docker check_base_files test_check_headers check_headers check_trailing_whitespace generate_docs ## Run all linters and update documentation
 
+.PHONY: check
+check: check_shell check_python check_golang check_terraform check_docker check_base_files test_check_headers check_headers check_trailing_whitespace
+
 # The .PHONY directive tells make that this isn't a real target and so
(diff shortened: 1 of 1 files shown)


morgante:
I think we should have a dedicated CI make task instead of relying on "all" for CI.

Maybe `make test`?

adrienthebo:
For the sake of discussion, what's the intent of the default `make`/`make all` target?

morgante:
`make` should essentially build everything I need as a developer: generate docs, generate root module, run tests, etc.

aaron-lane:
That sounds good. We will need to backport to terraform-google-module-template.

Makefile
```diff
@@ -23,6 +23,9 @@ DOCKER_REPO_BASE_KITCHEN_TERRAFORM := ${DOCKER_ORG}/cft/kitchen-terraform:${DOCK
 
 all: check_shell check_python check_golang check_terraform check_docker check_base_files test_check_headers check_headers check_trailing_whitespace generate_docs ## Run all linters and update documentation
```
morgante: Should we make all reference check instead of repeating all checks in both spots?

adrienthebo:
I need a `check` target for terraform-google-vpn so I've updated this functionality in the shared module template: https://github.com/terraform-google-modules/terraform-google-module-template/pull/7

morgante:
@aaron-lane Can you update this from the module template?

morgante:
@aaron-lane ping

aaron-lane:
Fixed by #187.

PR #134, opened 2019-01-24
Break out deprecated app_engine arg into its own resource/module (v2.0)

nstogner:
The `app_engine` field in the `google_project` resource is deprecated.

Options:
1. Remove app engine functionality from project factory
2. Integrate the app engine resource into project factory

This PR is for option 2.

Option 2 requires removing the `app_engine` arg and replacing with more granular args. This is needed b/c of terraform's limitation in accessing nested map fields (`app_engine["feature_settings"]`). Note: Option 2 also requires the user to `terraform import` the new resource as a part of the upgrade path, ex:
```sh
terraform import module.project-factory.module.project-factory.module.app-engine.google_app_engine_application.app pf-ci-test-full-y2d5b2ae-0d6c
```


adrienthebo:
Noting for posterity that this is a backwards incompatible change since it changes the Project Factory interface.

Zest-nikunj:
@nstogner Whenever I try to run a terraform plan with the module, I get an error:
project-factory.google_project.main: "app_engine": this field cannot be set
Any suggestions?

mikeraimondi:
I am also getting the error "project-factory.google_project.main: "app_engine": this field cannot be set" with v2.0.0 of the Google provider.

morgante:
Replaced with #144.

Zest-nikunj:
> I am also getting the error "project-factory.google_project.main: "app_engine": this field cannot be set" with v2.0.0 of the Google provider.

I had to comment the app_engine variable, https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L93
to make it work.

PR #141, opened 2019-02-20
Fix a race condition when creating a new G Suite group

thefirstofthe300:
When a new G Suite group is created to manage a project, `core_project_factory` would attempt to assign the group IAM permissions before the group was finished being created by the `gsuite_enabled` module. To fix this condition, an implicit dependency was added to the Terraform using the email attribute from the `gsuite_group` resource.

Also, the `google_compute_default_service_account` resource depends on the Compute Engine API being enabled so it is possible for the fetch of the data resource to fail because it attempts to query the Compute Engine API before it is fully enabled. Adding an explicit dependency on the services being enabled fixes this issue.

### DIFF
--- a/modules/gsuite_enabled/main.tf
+++ b/modules/gsuite_enabled/main.tf
@@ -27,8 +27,6 @@ resource "gsuite_group_member" "service_account_sa_group_member" {
   group = "${var.sa_group}"
   email = "${module.project-factory.service_account_email}"
   role  = "MEMBER"
-
-  depends_on = ["module.project-factory"]
 }
 
 /*****************************************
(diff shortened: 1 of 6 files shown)


modules/gsuite_enabled/main.tf
```diff
@@ -67,7 +65,7 @@ resource "gsuite_group_member" "api_s_account_api_sa_group_member" {
 module "project-factory" {
   source = "../core_project_factory/"
 
-  group_email                 = "${module.gsuite_group.email}"
+  group_email                 = "${gsuite_group.group.email}"
```
morgante: I don't think this logic will work for cases where `create_group = false`. I think we might be better off with an explicit `depends_on` to solve this issue.

thefirstofthe300: Good catch. I just tested that scenario and you are correct about there being a problem when `create_group = false`.

Unfortunately, Terraform doesn't support depending on a module yet so that's not an option to fix the race. The `gsuite_group` resource needs to be referenced at some point in the graph to create the implicit dependency.

That being said, I came up with a bit of a hack to get the group definition into the graph or to use the provided email if `create_group` is set to false. It appears to be working as expected.

PTAL.

PR #147, opened 2019-02-22
Fix regression in v1.1.0

thefirstofthe300:
Don't create a resource in the core project factory module that pulls in the gsuite provider.

### DIFF
--- a/modules/gsuite_enabled/main.tf
+++ b/modules/gsuite_enabled/main.tf
@@ -38,7 +38,17 @@ module "gsuite_group" {
   domain = "${var.domain}"
   name   = "${local.group_name}"
   org_id = "${var.org_id}"
-  create_group = "${var.create_group}"
+}
+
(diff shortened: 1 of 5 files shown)


modules/gsuite_enabled/main.tf
```diff
@@ -38,7 +38,17 @@ module "gsuite_group" {
...
+
+/******************************************
+  Gsuite Group Configuration
+ *****************************************/
+resource "gsuite_group" "group" {
+  count = "${var.create_group ? 1 : 0}"
+
+  description = "${var.name} project group"
+  email       = "${module.gsuite_group.email}"
+  name        = "${var.name}"
```
adrienthebo: Should this be `local.group_name`? `var.name` is the name of the project itself, right?

thefirstofthe300: Done.

PR #150, opened 2019-02-25
Improve readability of the README.

woogie-wolgemuth:
There's a tendency to just copy-paste the first visible code block when
reading docs, but that means users will miss version pinning the Google
providers. This updates the example block to the minimum viable config
(adds the provider blocks for both google and google-beta, with version
pinning).


morgante:
I'm not sure we should include the provider block in all modules, since it's a pre-requisite for all anyways.

PR #151, opened 2019-02-25
Introduce setup-sa-folder.sh

NFollett89:
- Add `setup-sa-folder.sh` which sets up the Seed SA with most permissions bound to a Folder instead of the Org.
- Rename `setup-sa.sh` to `setup-sa-org.sh`
- Update `README.md` to reflect changes to Helper scripts


NFollett89:
This is WIP for now - Jeff is going to validate that this works as intended before we move forward.

NFollett89:
Going to work on this in the Fork, will open PR when it is in better shape.

PR #153, opened 2019-02-27
Stablize default service account in plan

thefirstofthe300:
The default service account data resource currently uses a depends_on
flag added to prevent a race condition in
https://github.com/terraform-google-modules/terraform-google-project-factory/pull/141

Due to the way that Terraform refreshes data resources, Terraform thinks
that the data resource has changed when in actuality it hasn't:
https://github.com/hashicorp/terraform/issues/11806#issuecomment-279108107

By changing to use a null data resource that interpolates the default
service account email, the data resource will only change when the project
number does.

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -135,10 +135,10 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
 /******************************************
   Default compute service account retrieval
  *****************************************/
-data "google_compute_default_service_account" "default" {
-  project = "${google_project.main.id}"
-
-  depends_on = ["google_project_service.project_services"]
+data "null_data_source" "default_service_account" {
+  inputs = {
+    email = "${google_project.main.number}-compute@developer.gserviceaccount.com"
+  }
 }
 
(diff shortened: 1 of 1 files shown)


modules/core_project_factory/main.tf
```diff
@@ -135,26 +135,26 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
...
   }
 
   triggers {
-    default_service_account = "${data.google_compute_default_service_account.default.id}"
+    default_service_account = "${data.null_data_source.default_service_account.outputs["email"]}"
     activated_apis          = "${join(",", var.activate_apis)}"
   }
 
-  depends_on = ["google_project_service.project_services", "data.google_compute_default_service_account.default"]
+  depends_on = ["google_project_service.project_services", "data.null_data_source.default_service_account"]
```
adrienthebo: Since we're using the value of `data.null_data_source.default_service_account.outputs["email"]`, is this explicit dependency on the data source necessary?

thefirstofthe300: Good point. Removed.

PR #154, opened 2019-03-01
Allow project name to be changed separately from project ID

### LINKED ISSUE #132: Allow updating project name without recreating project
Currently, any changes to the project name updates the project ID, requiring a full recreation of the project. This is causing us significant trouble as we roll out new governance and standards changes across the organization. In local testing, adding `ignore_changes = ["project_id"]` to the google_project's lifecycle fixes this.

I'm happy to submit a PR if this is an acceptable direction.

thefirstofthe300:
Fixes #132.

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -26,18 +26,17 @@ resource "random_id" "random_project_id_suffix" {
  *****************************************/
 locals {
   group_id               = "${var.manage_group ? format("group:%s", var.group_email) : ""}"
-  project_id             = "${google_project.main.project_id}"
-  project_number         = "${google_project.main.number}"
+  base_project_id        = "${var.project_id == "" ? var.name : var.project_id}"
   project_org_id         = "${var.folder_id != "" ? "" : var.org_id}"
   project_folder_id      = "${var.folder_id != "" ? var.folder_id : ""}"
-  temp_project_id        = "${var.random_project_id ? format("%s-%s",var.name,random_id.random_project_id_suffix.hex) : var.name}"
+  temp_project_id        = "${var.random_project_id ? format("%s-%s",local.base_project_id,random_id.random_project_id_suffix.hex) : local.base_project_id}"
   s_account_fmt          = "${format("serviceAccount:%s", google_service_account.default_service_account.email)}"
-  api_s_account          = "${format("%s@cloudservices.gserviceaccount.com", local.project_number)}"
+  api_s_account          = "${format("%s@cloudservices.gserviceaccount.com", google_project.main.number)}"
   api_s_account_fmt      = "${format("serviceAccount:%s", local.api_s_account)}"
   gke_shared_vpc_enabled = "${var.shared_vpc != "" && contains(var.activate_apis, "container.googleapis.com") ? "true" : "false"}"
-  gke_s_account          = "${format("service-%s@container-engine-robot.iam.gserviceaccount.com", local.project_number)}"
+  gke_s_account          = "${format("service-%s@container-engine-robot.iam.gserviceaccount.com", google_project.main.number)}"
   gke_s_account_fmt      = "${local.gke_shared_vpc_enabled ? format("serviceAccount:%s", local.gke_s_account) : ""}"
-  project_bucket_name    = "${var.bucket_name != "" ? var.bucket_name : format("%s-state", var.name)}"
+  project_bucket_name    = "${var.bucket_name != "" ? var.bucket_name : format("%s-state", local.temp_project_id)}"
   create_bucket          = "${var.bucket_project != "" ? "true" : "false"}"
 
   shared_vpc_users = "${compact(list(local.group_id, local.s_account_fmt, local.api_s_account_fmt, local.gke_s_account_fmt))}"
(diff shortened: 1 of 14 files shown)


thefirstofthe300:
> The `project_id` interface needs to be added to the `gsuite_enabled` module.

I think you are referring to [this](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/154/files#diff-6d3eb8e1c4352b607a26fd0ca4cda6a6R75) change?

> We should test these changes against the minimal suite converged from the master branch. Are there any unexpected consequences just from updating? Can the name be changed without destroying resources?

https://gist.github.com/thefirstofthe300/d72f2d563196e1fce1ab36b734e64fdb

The minimal test converges properly. The full test suite recreates the project because the project ID changes. The precondition being rebuilt is due to an old container version being used and causing the shell command to fail. I'm going to be submitting a PR tomorrow to update the container and make targets since the targets themselves fail right now.

aaron-lane:
@thefirstofthe300:

Apologies: I missed the changes to `gsuite_enabled`. :blush: 

With respect to testing the full suite, can you verify the impact of changing _only_ the name?

modules/core_project_factory/main.tf
```diff
@@ -26,18 +26,17 @@ resource "random_id" "random_project_id_suffix" {
...
-  project_id             = "${google_project.main.project_id}"
-  project_number         = "${google_project.main.number}"
   project_org_id         = "${var.folder_id != "" ? "" : var.org_id}"
   project_folder_id      = "${var.folder_id != "" ? var.folder_id : ""}"
-  temp_project_id        = "${var.random_project_id ? format("%s-%s",var.name,random_id.random_project_id_suffix.hex) : var.name}"
+  project_id             = "${var.project_id == "" ? var.name : var.project_id}"
+  temp_project_id        = "${var.random_project_id ? format("%s-%s",local.project_id,random_id.random_project_id_suffix.hex) : local.project_id}"
   s_account_fmt          = "${format("serviceAccount:%s", google_service_account.default_service_account.email)}"
-  api_s_account          = "${format("%s@cloudservices.gserviceaccount.com", local.project_number)}"
+  api_s_account          = "${format("%s@cloudservices.gserviceaccount.com", google_project.main.number)}"
```
morgante: Why can't this continue to reference `project_number`?

modules/core_project_factory/main.tf
```diff
@@ -26,18 +26,17 @@ resource "random_id" "random_project_id_suffix" {
  *****************************************/
 locals {
   group_id               = "${var.manage_group ? format("group:%s", var.group_email) : ""}"
-  project_id             = "${google_project.main.project_id}"
-  project_number         = "${google_project.main.number}"
```
morgante: Why drop `local.project_number`?

modules/core_project_factory/main.tf
```diff
@@ -26,18 +26,17 @@ resource "random_id" "random_project_id_suffix" {
...
   project_org_id         = "${var.folder_id != "" ? "" : var.org_id}"
   project_folder_id      = "${var.folder_id != "" ? var.folder_id : ""}"
-  temp_project_id        = "${var.random_project_id ? format("%s-%s",var.name,random_id.random_project_id_suffix.hex) : var.name}"
+  project_id             = "${var.project_id == "" ? var.name : var.project_id}"
```
morgante: I think we should continue to have `project_id` reference the actual project created.

We can move the conditional check into the project ID itself.

modules/core_project_factory/main.tf
```diff
@@ -104,7 +103,7 @@ resource "google_resource_manager_lien" "lien" {
 resource "google_project_service" "project_services" {
   count = "${length(var.activate_apis)}"
 
-  project = "${local.project_id}"
+  project = "${google_project.main.project_id}"
```
morgante: We should keep `local.project_id` unless there's a reason we can't, similarly for all cases of this replacement.

modules/core_project_factory/main.tf
thefirstofthe300: This local doesn't necessarily refer to the value that will be used to set the project's ID. I'll rename these variables to make the naming a bit more sane.

modules/core_project_factory/main.tf
thefirstofthe300: The local here is a level of indirection that is unnecessary, IMHO.

modules/core_project_factory/main.tf
thefirstofthe300: The net effect of this change is a no-op. It removes a level of indirection since `local.project_number` is referring to `google_project.main.number`. I see no purpose for continuing to use `local.project_number`. It makes things a bit more difficult to read so I would much prefer to remove it.

modules/core_project_factory/main.tf
thefirstofthe300: I am of the opinion that the project_id local should be removed or repurposed to make the project_id generation more clear. It also adds a layer of redirection that seems unnecessary so my comments on dropping the `project_number` local applies as well.

modules/core_project_factory/main.tf
morgante: It's easier to manage this in one place than every place.

test/integration/full/controls/project-factory.rb
```diff
@@ -51,10 +63,6 @@
       end
     end
 
-    it "includes the Google App Engine API service account user" do
```
morgante: @thefirstofthe300 Please do not introduce completely unrelated changes into a different PR. Is this somehow related to the other functionality in this PR and I missed it?

PR #98, opened 2019-01-10
[WIP] Fix networking setup

morgante:
Fixes:
- #97

Issue #97, opened 2019-01-10
Shared VPC granting is broken


morgante:
In #72, we messed up the logic for granting networkUser permission.

There are meant to be 3 different conditions.

1. No `shared_vpc` and no subnets => no grants
2. `shared_vpc` and no subnets => grant to the project (all networks/subnets)
3. `shared_vpc` and subnets => grant to the subnets (and not to the project)

leone145:
I have verified on our end that the fix listed does correct the condition of not adding the group_name to the project level if condition 3 is met. However group_name does not get cleaned up from the shared subnet on project destroy.

aaron-lane:
Fixed by #164.

PR #164, opened 2019-03-07
Fix shared VPC IAM bindings

glarizza:
This commit addresses issue #97
(https://github.com/terraform-google-modules/terraform-google-project-factory/issues/97)
and updates the logic around IAM bindings with regard to shared VPC
subnets. The logic is as follows:

1. If `var.shared_vpc` and `var.shared_vpc_subnets` are empty no
   bindings are mad
2. If `var.shared_vpc` is set but no subnets are provided with
   `var.shared_vpc_subnets` then the IAM bindings are set at the Host
   Project
3. If `var.shared_vpc` is set and `var.shared_vpc_subnets` contains
   subnets then the IAM bindings are granted on the subnetworks
   themselve

This commit updates the logic used to calculate the Host Project
bindings based on scenario 3 above. The tests have also been modified to
ensure that those bindings AREN'T set.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,8 @@ determining that location is as follows:
 | lien | Add a lien on the project to prevent accidental deletion | string | `"false"` | no |
 | name | The name for the project | string | n/a | yes |
 | org\_id | The organization ID. | string | n/a | yes |
-| random\_project\_id | Enables project random id generation | string | `"false"` | no |
+| project\_id | If provided, the project uses the given project ID. Mutually exclusive with random_project_id being true. | string | `""` | no |
+| random\_project\_id | Enables project random id generation. Mutually exclusive with project_id being non-empty. | string | `"false"` | no |
 | sa\_role | A role to give the default Service Account for the project (defaults to none) | string | `""` | no |
 | shared\_vpc | The ID of the host project which hosts the shared VPC | string | `""` | no |
 | shared\_vpc\_subnets | List of subnets fully qualified subnet IDs (ie. projects/$project_id/regions/$region/subnetworks/$subnet_id) | list | `<list>` | no |
@@ -137,6 +138,7 @@ determining that location is as follows:
 | project\_bucket\_self\_link | Project's bucket selfLink |
 | project\_bucket\_url | Project's bucket url |
 | project\_id |  |
+| project\_name |  |
 | project\_number |  |
 | service\_account\_display\_name | The display name of the default service account |
 | service\_account\_email | The email of the default service account |
--- a/README.md
+++ b/README.md
@@ -86,6 +86,20 @@ The roles granted are specifically:
   - `compute.networkUser` on host project or specified subnets
   - `storage.admin` on `bucket_name` GCS bucket
 
+### Shared VPC subnets and IAM permissions
+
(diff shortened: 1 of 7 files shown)


aaron-lane:
Closes #98.

morgante:
> We need at least 3 suites to ensure that all of the cases in #97 are tested. This should include additional examples and test fixtures.

Agreed, but it would be good if we can merge this sooner rather than later. (ie. I don't think new tests should be blocking.)

test/integration/shared_vpc_no_subnets/controls/gcloud.rb
```diff
@@ -0,0 +1,73 @@
...
+group_email                 = attribute('group_email')
+project_id                  = attribute('project_id')
+project_number              = attribute('project_number')
+service_account_email       = attribute('service_account_email')
+shared_vpc                  = attribute('shared_vpc')
+
+control 'project-factory-shared-vpc' do
+  title "Project Factory shared VPC"
+
+  only_if { !(shared_vpc.nil? || shared_vpc == '') }
```
adrienthebo: Cleanup/nonblocker: this `only_if` expression stems from when I wrote a single integration test suite that ran with different parameters to test different modes. Since this test requires a shared VPC I think we can drop this check.

It may be better to follow up on this in a later PR, so if we want a quick turnaround resolve this issue and ping me; I'll file a ticket to follow up on this.

test/integration/shared_vpc_no_subnets/controls/gcloud.rb
```diff
@@ -0,0 +1,73 @@
...
+    describe "roles/compute.networkUser" do
+      it "includes the project service account in the roles/compute.networkUser IAM binding" do
+        expect(bindings).to include(
+          members: including("serviceAccount:#{service_account_email}"),
+          role: "roles/compute.networkUser",
+        )
+      end
+
+      it "includes the group email in the roles/compute.networkUser IAM binding" do
+        if group_email.nil? || group_email.empty?
```
adrienthebo: Cleanup: I think this conditional can be dropped since `group_email` is required. As per the earlier comment we can defer this work in order to get this PR turned around.

glarizza:
Okay, I've reverted the commit back to the fix + updated existing test, and the README updates that came out of running `make`. We can add the test for scenario 2 in a later PR. Tests are going green now so we should be good to go - let me know if there's anything additional that's needed.

README.md
```diff
@@ -86,49 +86,66 @@ The roles granted are specifically:
   - `compute.networkUser` on host project or specified subnets
   - `storage.admin` on `bucket_name` GCS bucket
 
+### Shared VPC subnets and IAM permissions
+
+A service project's access to shared VPC networks is controlled via the
+`roles/compute.networkUser` role and the location to where that role is
+assigned. If that role is assigned to the shared VPC host project, then the
+service project will have access to ALL shared vpc subnetworks. If that role
```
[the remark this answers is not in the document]
glarizza: Added to the original commit

glarizza:
@aaron-lane That makes sense why it was recommending those changes - updated `terraform-docs`, added you change, and force-pushed

PR #166, opened 2019-03-12
Add test for shared VPC IAM behavior

glarizza:
This commit adds tests surrounding the shared VPC IAM behavior.
Specifically, we're testing whether the `compute.networkUser` role is
added to the Host Project if `var.shared_vpc` is passed but
`var.shared_vpc_subnets` is not.

### DIFF
--- a/test/fixtures/shared_vpc_no_subnets/main.tf
+++ b/test/fixtures/shared_vpc_no_subnets/main.tf
@@ -1,0 +1,59 @@
+/**
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+provider "google" {
+  version = "~> 2.1"
+}
+
+provider "google-beta" {
+  version = "~> 2.1"
+}
+
+provider "gsuite" {
+  credentials             = "${file(var.credentials_path)}"
+  impersonated_user_email = "${var.gsuite_admin_account}"
+
+  oauth_scopes = [
+    "https://www.googleapis.com/auth/admin.directory.group",
+    "https://www.googleapis.com/auth/admin.directory.group.member",
+  ]
+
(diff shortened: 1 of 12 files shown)


test/integration/shared_vpc_no_subnets/inspec.yml
```diff
@@ -0,0 +1,34 @@
...
+  - name: project_number
+    required: true
+    type: string
+
+  - name: service_account_email
+    required: true
+    type: string
+
+  - name: shared_vpc
+    required: false
```
aaron-lane: Attributes should be required to ensure that Terraform outputs match in naming.

```suggestion
    required: true
```

test/integration/shared_vpc_no_subnets/inspec.yml
```diff
@@ -0,0 +1,34 @@
...
+    default: null
```
aaron-lane: A default value is not needed as the value should always be provided by Terraform.

```suggestion
```

test/fixtures/shared_vpc_no_subnets/main.tf
```diff
@@ -0,0 +1,65 @@
...
+  name              = "pf-ci-test-nosubnets-${random_string.suffix.result}"
+  project_id        = "pf-ci-test-nosubnets-${random_string.suffix.result}"
+  random_project_id = false
+  domain            = "${var.domain}"
+  org_id            = "${var.org_id}"
+  folder_id         = "${var.folder_id}"
+  billing_account   = "${var.billing_account}"
+  create_group      = "true"
+  group_role        = "${var.group_role}"
+  group_name        = "pf-secondary-group"
```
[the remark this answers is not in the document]
glarizza: Yeah, that works out well. I used this pattern but adjusted it to ensure it would work for macOS too

glarizza:
@aaron-lane Made the changes you recommended (and one coming from `shellcheck`), pushed, and tests are passing. It's ready for another review when you have a moment

glarizza:
Updated the PR to add a test for the compute.networkUser role at the subnetwork level as part of the `full` testing scenario that passes in `var.shared_vpc` and `var.shared_vpc_subnets`

glarizza:
Updated the PR to add a test looping through all subnets in all regions on the shared VPC host project to ensure that compute.networkUser bindings aren't set at the subnetwork level when only `var.shared_vpc` is passed in without passing in any subnets via `var.shared_vpc_subnets`

glarizza:
Marking this as WIP for now so I can unify the behavior between the two random strings based on #169 that was just merged. I'll update when it's ready for another review.

glarizza:
Okay, I've updated the test to use the variable created in #169 to ensure all test resources are appended with the same random string. Changes have been pushed, tests are passing, and Ideally this will be the final review @aaron-lane  :)

PR #171, opened 2019-03-17
feat: add pre-commit to automatically format and create docs

gabel0287:
### Why

I find pre-commit useful in that it won't let me commit if terraform is not properly formatted or docs need to be updated.


morgante:
Thanks for the contribution, can you add a license header? https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/test/boilerplate/boilerplate.yaml.txt

PR #172, opened 2019-03-18
Fix bug in gsuite_group module that requires organizationViewer role

sosimon:
* `roles/organizationViewer` role should not be necessary if domain is specified.
* Error occurs when project factory is scoped to the folder level (i.e. no org-level permissions, only folder-level permissions)

```
data.google_organization.org: Refreshing state...
       Error: Error refreshing state: 1 error(s) occurred:
       * module.project-factory.module.gsuite_group.data.google_organization.org: 1 error(s) occurred:
       * module.project-factory.module.gsuite_group.data.google_organization.org: data.google_organization.org: Error reading Organization Not Found : <org-id-redacted>: googleapi: Error 403: The caller does not have permission, forbidden
```

The fix disables the `data.organization` lookup when the variable `domain` is specified.

### DIFF
--- a/modules/gsuite_group/main.tf
+++ b/modules/gsuite_group/main.tf
@@ -15,13 +15,15 @@
  */
 
 locals {
-  domain = "${var.domain != "" ? var.domain : data.google_organization.org.domain}"
-  email  = "${format("%s@%s", var.name, local.domain)}"
+  domain_list = "${concat(data.google_organization.org.*.domain, list("dummy"))}"
+  domain      = "${var.domain == "" ? element(local.domain_list, 0) : var.domain}"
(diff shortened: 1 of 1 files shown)


morgante:
We should also add a test suite to make sure this doesn't regress again.

sosimon:
@morgante @aaron-lane 
I did some investigation and I don't believe we need an additional test suite. Our existing test suite would have failed (as we would expect) if the CI service account did not have `resourcemanager.organizationViewer` at the org-level (it has it right now, but it shouldn't; it's not defined in our test_fixtures). I created a separate bug to track the permissions cleanup on our CI service account (b/129068671)

PR #178, opened 2019-04-01
remove permission check for resourcemanager.organization.get

sosimon:
This org-level permission is not required in most cases - it is only required for the `gsuite_group` module when "domain" is not specified. Enhancement to add this conditional verification to the `gsuite_group` module will be handled in a subsequent, separate PR.

### DIFF
--- a/test/scripts/preconditions/test_preconditions.py
+++ b/test/scripts/preconditions/test_preconditions.py
@@ -107,21 +107,11 @@ class TestRequirements(unittest.TestCase):
 
 
 class TestOrgPermissions(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


test/scripts/preconditions/test_preconditions.py
```diff
@@ -107,21 +107,11 @@ def test_empty_required(self):
 
 
 class TestOrgPermissions(unittest.TestCase):
-    def test_base_permissions(self):
```
aaron-lane: Let's retain this test and assert that the base permissions are empty. This could change again in the future.

PR #180, opened 2019-04-03
Add Shared VPC example and fix dependencies

morgante:
This PR:

- Adds an example of combining the project factory and shared VPC modules to create a host project
- Fixes https://github.com/terraform-google-modules/terraform-google-network/issues/15 by making the project ID output depend on project service activation

### DIFF
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -1,0 +1,80 @@
+/**
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


examples/shared_vpc/main.tf
```diff
@@ -0,0 +1,80 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+locals {
+  credentials_file_path = "${var.credentials_path}"
```
aaron-lane: This local is redundant.

modules/core_project_factory/outputs.tf
```diff
@@ -19,7 +19,7 @@ output "project_name" {
 }
 
 output "project_id" {
-  value = "${google_project.main.project_id}"
+  value = "${element(concat(google_project_service.project_services.*.project, list(local.project_id)), 0)}"
```
aaron-lane: `local.project_id` is undefined.

examples/shared_vpc/main.tf
morgante: It's the existing convention for examples.

Issue #152, opened 2019-02-27
GCP subnet share conditions not working correctly


sandhyajob:
Based on the default 3 conditions
1. No `shared_vpc` and no subnets => no grants
2. `shared_vpc` and no subnets => grant to the project (all networks/subnets)
3. `shared_vpc` and subnets => grant to the subnets (and not to the project)

If the subnet id get messed up ( fat finger) while attempting case 3, project factory ends up sharing the entire set of networks in the host project VPC to the service project. Which is a serious security issue. Ideally it should follow step 1 and should not grant any thing and error out.

leone145:
When variable vpc_networks is incorrectly formed the failure happens and leaves all subnets shared
Correct formatting and functional	
shared_vpc_subnets: '["projects/vpc-abc/regions/europe-west3/subnetworks/subnet1"]'


Incorrect formatting and error observed below
shared_vpc_subnets: '["projects/vpc-abc/regions/europe-west3/subnet1"]'

* google_compute_subnetwork_iam_member.service_account_role_to_vpc_subnets: Error retrieving IAM policy for Compute Subnetwork mcktech-non-prod-host/europe-west3/projects: googleapi: Error 404: The resource 'projects/non-prod-host/regions/europe-west3/subnetworks/projects' was not found, notFound
	
* module.core-factory.google_compute_subnetwork_iam_member.apis_service_account_role_to_vpc_subnets: 1 error(s) occurred:
	

	
* google_compute_subnetwork_iam_member.apis_service_account_role_to_vpc_subnets: Error retrieving IAM policy for Compute Subnetwork mcktech-non-prod-host/europe-west3/projects: googleapi: Error 404: The resource 'projects/non-prod-host/regions/europe-west3/subnetworks/projects' was not found, notFound

leone145:
When variable vpc_networks is incorrectly formed the failure happens and leaves all subnets shared
Correct formatting and functional
shared_vpc_subnets: '["projects/vpc-abc/regions/europe-west3/subnetworks/subnet1"]'

Incorrect formatting and error observed below
shared_vpc_subnets: '["projects/vpc-abc/regions/europe-west3/subnet1"]'

[...]

PR #194, opened 2019-05-08
GCP subnet share conditions not working correctly

kopachevsky:
Issue  #152

Ready for review


aaron-lane:
Fixes #152.

modules/core_project_factory/main.tf
```diff
@@ -70,6 +70,28 @@ EOD
...
+resource "null_resource" "shared_vpc_subnet_invalid_name" {
+  count = "${length(var.shared_vpc_subnets)}"
+
+  triggers = {
+    name = "${replace(var.shared_vpc_subnets[count.index],
+    "/(https://www.googleapis.com/compute/v1/)?projects/[a-z0-9-]+/regions/[a-z0-9-]+/subnetworks/[a-z0-9-]+/", "") }"
+  }
+}
+
+resource "null_resource" "check_if_shared_vpc_subnets_contais_items_with_invalid_name" {
```
aaron-lane:
```suggestion
resource "null_resource" "check_if_shared_vpc_subnets_contains_items_with_invalid_name" {
```

test/integration/full/controls/shared-vpc.rb
```diff
@@ -88,28 +88,41 @@
...
+      "name" => shared_vpc_subnet_name_01,
+      "region" => shared_vpc_subnet_region_01
+    },
+    {
+      "name" => shared_vpc_subnet_name_01,
+      "region" => shared_vpc_subnet_region_01
+    }
+  ]
+
+  SHARED_VPC_SUBNETS_DATA_SET.each do |subnet|
```
aaron-lane: Test cases must not be generated through metaprogramming; this practice can lead to false positives if the logic driving the generation happens to change.

kopachevsky: I agree, but how would you minimize code duplication in this case? basically we need to do same code against 2 different networks

aaron-lane: Reducing code duplication in test cases is less of a priority than creating clear test cases. Duplicating logic in this scenario is totally acceptable.

kopachevsky:
@aaron-lane 
> Can you please add a CHANGELOG entry and provide a [gist](https://gist.github.com) which contains the output of a Terraform process when it fails due to an invalid subnet being passed to `shared_vpc_subnets`?

Added CHANGELOG
Regarding gist, it can take to much time execute locally full test without G Suite, so I've added bad sunet name in previous commit and got:
       
```
       1 error(s) occurred:
       
       * module.project-factory.module.project-factory.null_resource.check_if_shared_vpc_subnets_contains_items_with_invalid_name: Error running command 'false': exit status 1. Output: 
```
       
on Concourse  build 422, should I copy/paste full terraform output from there for Full test suite?  I think it could contain sensitive info

PR #199, opened 2019-05-13
integration test should fail if group_name is absent in email

alexkonkin:


alexkonkin:
Please don't review this PR during the first iteration, I would like to have it checked by the CI system

PR #201, opened 2019-05-15
Fabric project submodule initial import

ludoo:
As agreed, this is the initial import of our project module, as a submodule.


modules/fabric-project/variables.tf
```diff
@@ -0,0 +1,110 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "parent_id" {
+  description = "Id of the resource under which the folder will be placed."
+}
+
+variable "parent_type" {
+  description = "Type of the parent reosurce, defaults to organization."
```
morgante: Spelling.

modules/fabric-project/variables.tf
```diff
@@ -0,0 +1,110 @@
...
+
+variable "prefix" {
+  description = "Prefix used to generate project id and name"
+}
+
+variable "name" {
+  description = "Project name and id suffix."
+}
+
+variable "billing_account_id" {
```
morgante: Let's standardize on the same name for shared variables with the standard module:
- billing_account
- activate_apis

ludoo: I'm not too keen on the `activate_apis` name, as it masks the actual resource name. But can see the benefits of standardizing, lemme fix that.

morgante:
@ludoo LGTM, merged.

In a future update, we should consider whether we want to offer an easy path for moving between different versions of the module.

PR #207, opened 2019-05-22
project_factory: added bucket_location property

alexkonkin:


modules/core_project_factory/variables.tf
```diff
@@ -116,7 +116,7 @@ variable "bucket_name" {
 
 variable "bucket_location" {
   description = "The location for a GCS bucket to create (optional)"
-  default     = "US"
+  default     = ""
```
morgante: Why are you changing this default? Blank isn't a valid default.

Issue #102, opened 2019-01-10
Precondition step fails if no org_id supplied


angstwad:
```
  File ".terraform/modules/c158d4e24da0dc3f5e10ff9e6d98ef82/scripts/preconditions/preconditions.py", line 425, in <module>
    retcode = main(sys.argv)
  File ".terraform/modules/c158d4e24da0dc3f5e10ff9e6d98ef82/scripts/preconditions/preconditions.py", line 406, in main
    results.append(validator.validate(credentials))
  File ".terraform/modules/c158d4e24da0dc3f5e10ff9e6d98ef82/scripts/preconditions/preconditions.py", line 125, in validate
    resource = "organizations/" + self.org_id
TypeError: can only concatenate str (not "NoneType") to str
```

angstwad:
`org_id` as an argument was supplied with a value of an empty string, which was eval'ed as `None`

morgante:
I believe we're moving to making `org_id` a required variable.

adrienthebo:
`org_id` became a required variable in acfad021718db2a184972cdc594f418e255317f8 as part of the preconditions script requirements. Making the `org_id` required also avoids some permissions issue; without it I believe that the Project Factory requires `roles/resourcemanager.organizationViewer`. Since the org ID should be readily available and making it required simplifies the module, I'm inclined to keep this as a required.

Is there a compelling case for making `org_id` optional that we haven't accounted for?

ocervell:
@adrienthebo Some more details in #20.
tl;dr: sometimes sub-teams want to create projects and have restricted access (no org viewer role granted to their group / Terraform service account). Ideally, creating a project should just require the 'projectCreator' role under the folder / org they're spinning it up under.
Why do we need `org_id` in the pre-requisites script ?

morgante:
@ocervell Can you clarify whether the issue is with requiring the `org_id` attribute or requiring the Organization Viewer role?

I'd think the first requirement is fairly innocuous but agree we can/should drop dependencies on Org Viewer.

ocervell:
@morgante yes I meant Org Viewer, when creating a project under a folder Organization Viewer should not be required (but Folder Viewer might be)

aaron-lane:
This appears to have been solved with v2.1.3, specifically #178. @angstwad, @ocervell can you please confirm?

ocervell:
@aaron-lane yep, that should fix it since the precondition script was the one failing.

PR #210, opened 2019-05-23
Added check for org_id to precondition script

ingwarr:
Precondition step fails if no org_id supplied #102


modules/core_project_factory/scripts/preconditions/preconditions.py
```diff
@@ -431,18 +431,23 @@ def validators_for(opts, seed_project):
...
-            if len(result["unsatisfied"]) > 0:
-                retcode = 1
+            for validator in validators:
+                results.append(validator.validate(credentials))
+
+            retcode = 0
+            for result in results:
+                if len(result["unsatisfied"]) > 0:
+                    retcode = 1
+        else:
```
akopachevskyy-globallogic: what about folder_id maybe it's ok to have folder_id but not nessesary org_id?

ingwarr: We need to check org_id. There is validate function(which checks permissions) in precondition script and it fails if org_id is None. 
".terraform/modules/c158d4e24da0dc3f5e10ff9e6d98ef82/scripts/preconditions/preconditions.py", line 125, in validate
    resource = "organizations/" + self.org_id
TypeError: can only concatenate str (not "NoneType") to str

morgante:
@ingwarr I've realized that `org_id` is already a require variable for the module, so we shouldn't need/add a separate check in `preconditions.py`.

PR #212, opened 2019-05-23
Allow personal credentials

### LINKED ISSUE #211: Enable personal credentials to be used as default application credentials
The `preconditions.py` script ends up requiring the use of a service account's credentials as it is looking up which APIs are enabled on the project hosting the service account. When using personal credentials, there is no project, so it fails in checking which APIs are enabled on the hosting project, as there is no hosting project.

bbrouwer:
Fixes #211

### DIFF
--- a/modules/core_project_factory/scripts/preconditions/preconditions.py
+++ b/modules/core_project_factory/scripts/preconditions/preconditions.py
@@ -404,13 +404,13 @@ def validators_for(opts, seed_project):
     Given a set of CLI options, determine which preconditions we need
     to check and generate corresponding validators.
     """
-    validators = [
-        BillingAccount(opts.billing_account),
-    ]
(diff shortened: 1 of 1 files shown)


modules/core_project_factory/scripts/preconditions/preconditions.py
```diff
@@ -405,10 +405,13 @@ def validators_for(opts, seed_project):
     to check and generate corresponding validators.
     """
     validators = [
-        SeedProjectServices(seed_project),
         BillingAccount(opts.billing_account),
```
morgante: I believe ordering is important here, because you can reach a bad state if the seed project doesn't have required APIs but we try to validate the billing account.

Could you make sure BillingAccount validators get appended after the SeedProject?

Otherwise LGTM.

PR #215, opened 2019-05-27
Some fixes for integration tests

ingwarr:
There was an issue related to trying variables usage before variables creation in test/ci-integration.sh

### DIFF
--- a/README.md
+++ b/README.md
@@ -331,22 +331,26 @@ Two test-kitchen instances are defined:
 1. Configure the [test fixtures](#test-configuration).
 2. Download a Service Account key with the necessary [permissions](#permissions)
    and put it in the module's root directory with the name `credentials.json`.
-3. Build the Docker containers for testing.
-    ```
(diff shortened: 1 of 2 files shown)


ingwarr:
All variables should be declared before `set -eu` because:
 -u: _Treat unset variables and parameters other than the special parameters ‘@’ or ‘*’ as an error when performing parameter expansion. An error message will be written to the standard error, and a non-interactive shell will exit._

So we have three variants: move invocation of `setup_environment` to the position before `set -eu`, remove `set -eu`(or better to say remove `-u`) or move `set -eu` to the position after the invocation of `setup_environment`

What do you think about this?

aaron-lane:
@ingwarr we should correct the variables which are not being set rather than masking the problem by changing the script.

ingwarr:
@aaron-lane After discussion with you, I corrected variables in README.md and removed commit with fixes to test/ci-integration

Issue #209, opened 2019-05-23
Add support for Terraform v0.12


thefirstofthe300:
A dry-run of my current use case for project factory yields the following error output:
```
Error: Unsupported block type

  on .terraform/modules/pg_seed_project.seed_project.pg_project/terraform-google-modules-terraform-google-project-factory-f68648a/modules/core_project_factory/main.tf line 49, in resource "null_resource" "preconditions":
  49:   triggers {

Blocks of type "triggers" are not expected here. Did you mean to define
argument "triggers"? If so, use the equals sign to assign it a value.


Error: Missing resource instance key

  on .terraform/modules/pg_seed_project.seed_project.pg_project/terraform-google-modules-terraform-google-project-factory-f68648a/modules/core_project_factory/main.tf line 288, in resource "google_storage_bucket_iam_member" "group_storage_admin_on_project_bucket":
 288:   bucket = "${google_storage_bucket.project_bucket.name}"

Because google_storage_bucket.project_bucket has "count" set, its attributes
must be accessed on specific instances.

For example, to correlate with indices of a referring resource, use:
    google_storage_bucket.project_bucket[count.index]
```
It would be nice to be able to use the CFT with Terraform v0.12, even if the CFT doesn't yet use all of the new features for v0.12.

dhduvall:
Here's another error related to 0.12 (from #214): when I run terraform init (terraform version 0.12.0) I get a pile of messages like this:

```
Initializing modules...
Downloading terraform-google-modules/project-factory/google 2.2.1 for project-factory...
- project-factory in .terraform/modules/project-factory/terraform-google-modules-terraform-google-project-factory-9e646bf
Downloading modules/gsuite_group for project-factory.gsuite_group...
Downloading modules/core_project_factory for project-factory.project-factory...
Downloading terraform-google-modules/network/google 0.6.0 for vpc...
- vpc in .terraform/modules/vpc/terraform-google-modules-terraform-google-network-1d3242b

Error: Module not found

The module address "modules/gsuite_group" could not be resolved.

If you intended this as a path relative to the current module, use
"./modules/gsuite_group" instead. The "./" prefix indicates that the address
is a relative filesystem path.


Error: Module not found

The module address "modules/core_project_factory" could not be resolved.

If you intended this as a path relative to the current module, use
"./modules/core_project_factory" instead. The "./" prefix indicates that the
address is a relative filesystem path.
```

[...]

PR #222, opened 2019-06-06
Enable Terraform 0.12 support while still working with 0.11

bbrouwer:
#209 Add support for Terraform 0.12


bbrouwer:
I have no idea how complete this solution is, but it does unblock my use of this module for both Terraform 0.11 and 0.12.

aaron-lane:
@bbrouwer thank you for the contribution! We're implementing complete support with #225. Can you test the associated branch to see if it solves your use case and comment on that pull request accordingly?

PR #231, opened 2019-06-20
app_engine example fixed with upgrade to 0.12 terraform

ingwarr:


examples/app_engine/variables.tf
```diff
@@ -14,14 +14,28 @@
...
+variable "location_id" {
+  description = "The location to serve the app from."
+  default     = "us-central"
 }
 
-variable "billing_account" {
-  description = "The ID of the billing account to associate this project with"
+variable "auth_domain" {
+  description = "The domain to authenticate users with when using App Engine's User API."
+  default     = "phoogle.net"
```
aaron-lane: We should only use default values if they are valid for users.

```suggestion
```

examples/app_engine/variables.tf
```diff
@@ -14,14 +14,27 @@
  * limitations under the License.
  */
 
-variable "admin_email" {
-  description = "Admin user email on Gsuite"
+variable "project_id" {
+  description = "The project to enable app engine on."
+  default     = "example-project"
```
aaron-lane: This invalid default value should also be removed.

```suggestion
```

PR #233, opened 2019-06-20
Pass through apis_authority

### LINKED ISSUE #232: apis_authority isn't being passed as an input to modules/core_project_factory
apis_authority argument isn't working.  apis_authority isn't being passed as an input to modules/core_project_factory from terraform-google-modules/project-factory/google

https://github.com/terraform-google-modules/terraform-google-project-factory/blob/v2.4.0/main.tf

morgante:
Fixes #232


morgante:
Lint is unrelated.

Issue #191, opened 2019-05-03
Document installation of preconditions.py requirements


chrislovecnm:
The README does not mention that you need to install the prereqs for precondions.py to run. Maybe we should mention it?

```console
modules/core_project_factory/scripts/preconditions/preconditions.py \
  --credentials_path "/Users/chlove/.config/gcloud/application_default_credentials.json" \
  --billing_account  my-account \
  --org_id my-id
Traceback (most recent call last):
  File "modules/core_project_factory/scripts/preconditions/preconditions.py", line 34, in <module>
    raise e
  File "modules/core_project_factory/scripts/preconditions/preconditions.py", line 25, in <module>
    import google.auth
ModuleNotFoundError: No module named 'google'
```

I know that I need to run pip3 with the requirements files, but not everyone will know that :)

morgante:
Good idea, let's add this to the docs.

Eventually, we would like to replace Python scripts with static Go binaries.

aaron-lane:
We could add `pip3` to the provisioner logic as well.

mengesb:
So potentially a quick fix for this would be adding the install specified in the `requirements.txt`:

```shell
if command -v python3 1>/dev/null; then
    BASEDIR="$(dirname "$0")"

    if command -v pip3 1>/dev/null; then
        exec "pip3" "install" "-r" "$BASEDIR/preconditions/requirements.txt"
    fi

    SCRIPT="$BASEDIR/preconditions/preconditions.py"
    exec "$SCRIPT" "$@"
else
    echo "Unable to check project-factory preconditions: python3 executable not in PATH" 1>&2
fi
```

Of course if you want to extend the requirements installation to produce an else for errors then by all means, I was considering this as a PR

aaron-lane:
@mengesb nice. Feel free to open a PR!

PR #239, opened 2019-07-01
Attempt to install python requirements

mengesb:
Addresses #191 

Signed-off-by: Brian Menges <mengesb@gmail.com>


mengesb:
Not sure what's up with `concourse-ci/integration-tests`, can't see what's failed, but maybe some can relay that information

Issue #47, opened 2018-10-09
Fails with shared VPC being specified using a reference


harpenden:
If I specify shared_vpc in a service project I receive following error during plan or apply. If I remove shared_vpc it works but creates a standalone project instead of a service project.

```
* module.service_project01.google_compute_subnetwork_iam_member.group_role_to_vpc_subnets:
google_compute_subnetwork_iam_member.group_role_to_vpc_subnets: value of 'count' cannot be computed
* module.service_project01.google_compute_shared_vpc_service_project.shared_vpc_attachment: google_compute_shared_vpc_service_project.shared_vpc_attachment: value of 'count' cannot be computed
* module.service_project01.google_project_iam_member.controlling_group_vpc_membership: google_project_iam_member.controlling_group_vpc_membership: value of 'count' cannot be computed
* module.service_project01.google_compute_subnetwork_iam_member.gke_shared_vpc_subnets: google_compute_subnetwork_iam_member.gke_shared_vpc_subnets: value of 'count' cannot be computed
* module.service_project01.google_compute_subnetwork_iam_member.service_account_role_to_vpc_subnets: google_compute_subnetwork_iam_member.service_account_role_to_vpc_subnets: value of 'count' cannot be computed
* module.service_project01.google_compute_subnetwork_iam_member.apis_service_account_role_to_vpc_subnets: google_compute_subnetwork_iam_member.apis_service_account_role_to_vpc_subnets: value of 'count' cannot be computed
* module.service_project01.google_project_iam_member.gke_host_agent: google_project_iam_member.gke_host_agent: value of 'count' cannot be computed
```

**Example resources:**
```

module "host_project01" {
  source            = "/modules/project-factory"
  random_project_id = "true"
  name              = "host-project-01"
  org_id            = "${var.organization_id}"
  billing_account   = "${var.billing_account}"
  activate_apis     = ["compute.googleapis.com"]
  credentials_path  = "${local.credentials_file_path}"
}

module "service_project01" {
  source            = "/modules/project-factory"
  random_project_id = "true"
  name              = "service-project01"
  org_id            = "${var.organization_id}"
  billing_account   = "${var.billing_account}"
  folder_id         = "${google_folder.prod.id}"
  shared_vpc        = "${module.host_project01.project_id}" # If I remove this it works but creates a standalone project.
  activate_apis     = ["compute.googleapis.com"]
  credentials_path  = "${local.credentials_file_path}"
  group_name        = "testgroup"
[...]

PR #244, opened 2019-07-10
Fixes for Fails with shared VPC being specified using a reference #47

ingwarr:
Fixes https://github.com/terraform-google-modules/terraform-google-project-factory/issues/47

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -43,7 +43,7 @@ locals {
     google_project.main.number,
   )
   api_s_account_fmt      = format("serviceAccount:%s", local.api_s_account)
-  gke_shared_vpc_enabled = var.shared_vpc != "" && contains(var.activate_apis, "container.googleapis.com") ? "true" : "false"
+  gke_shared_vpc_enabled = var.shared_vpc_enabled && contains(var.activate_apis, "container.googleapis.com")
   gke_s_account = format(
     "service-%s@container-engine-robot.iam.gserviceaccount.com",
     google_project.main.number,
@@ -173,7 +173,7 @@ resource "google_project_services" "project_services_authority" {
   Shared VPC configuration
  *****************************************/
 resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
-  count = var.shared_vpc != "" ? 1 : 0
+  count = var.shared_vpc_enabled ? 1 : 0
 
   host_project = var.shared_vpc
   service_project = google_project.main.project_id
@@ -283,7 +283,7 @@ resource "google_service_account_iam_member" "service_account_grant_to_group" {
   Account on shared VPC
  *****************************************************************************************************************/
 resource "google_project_iam_member" "controlling_group_vpc_membership" {
-  count = var.shared_vpc != "" && length(compact(var.shared_vpc_subnets)) == 0 ? local.shared_vpc_users_length : 0
+  count = var.shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) == 0 ? local.shared_vpc_users_length : 0
 
   project = var.shared_vpc
   role = "roles/compute.networkUser"
@@ -300,8 +300,7 @@ resource "google_project_iam_member" "controlling_group_vpc_membership" {
  *************************************************************************************/
 resource "google_compute_subnetwork_iam_member" "service_account_role_to_vpc_subnets" {
   provider = google-beta
-
-  count = var.shared_vpc != "" && length(compact(var.shared_vpc_subnets)) > 0 ? length(var.shared_vpc_subnets) : 0
+  count = var.shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) > 0 ? length(var.shared_vpc_subnets) : 0
 
   subnetwork = element(
     split("/", var.shared_vpc_subnets[count.index]),
@@ -325,8 +324,7 @@ resource "google_compute_subnetwork_iam_member" "service_account_role_to_vpc_sub
 resource "google_compute_subnetwork_iam_member" "group_role_to_vpc_subnets" {
   provider = google-beta
 
-  count = var.shared_vpc != "" && length(compact(var.shared_vpc_subnets)) > 0 && var.manage_group ? length(var.shared_vpc_subnets) : 0
-
+  count = var.shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) > 0 && var.manage_group ? length(var.shared_vpc_subnets) : 0
   subnetwork = element(
     split("/", var.shared_vpc_subnets[count.index]),
     index(
@@ -349,8 +347,7 @@ resource "google_compute_subnetwork_iam_member" "group_role_to_vpc_subnets" {
 resource "google_compute_subnetwork_iam_member" "apis_service_account_role_to_vpc_subnets" {
   provider = google-beta
 
-  count = var.shared_vpc != "" && length(compact(var.shared_vpc_subnets)) > 0 ? length(var.shared_vpc_subnets) : 0
-
+  count = var.shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) > 0 ? length(var.shared_vpc_subnets) : 0
   subnetwork = element(
     split("/", var.shared_vpc_subnets[count.index]),
     index(
@@ -442,9 +439,7 @@ resource "google_storage_bucket_iam_member" "api_s_account_storage_admin_on_proj
  *****************************************/
 resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
   provider = google-beta
-
   count = local.gke_shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) != 0 ? length(var.shared_vpc_subnets) : 0
-
   subnetwork = element(
     split("/", var.shared_vpc_subnets[count.index]),
     index(
@@ -471,11 +466,9 @@ resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
  *****************************************/
 resource "google_project_iam_member" "gke_host_agent" {
   count = local.gke_shared_vpc_enabled ? 1 : 0
-
   project = var.shared_vpc
   role = "roles/container.hostServiceAgentUser"
   member = local.gke_s_account_fmt
-
   depends_on = [
     google_project_service.project_services,
     google_project_services.project_services_authority,
--- a/modules/core_project_factory/variables.tf
+++ b/modules/core_project_factory/variables.tf
@@ -25,13 +25,14 @@ variable "group_role" {
 
 variable "lien" {
   description = "Add a lien on the project to prevent accidental deletion"
-  default     = "false"
-  type        = string
+  default     = false
+  type        = bool
 }
 
 variable "manage_group" {
   description = "A toggle to indicate if a G Suite group should be managed."
-  default     = "false"
+  type        = bool
+  default     = false
 }
 
 variable "project_id" {
@@ -41,7 +42,8 @@ variable "project_id" {
 
 variable "random_project_id" {
   description = "Enables project random id generation. Mutually exclusive with project_id being non-empty."
-  default     = "false"
+  type        = bool
+  default     = false
 }
 
 variable "org_id" {
@@ -73,7 +75,8 @@ variable "sa_role" {
 
 variable "apis_authority" {
   description = "Toggles authoritative management of project services."
-  default     = "false"
+  type        = bool
+  default     = false
 }
 
 variable "activate_apis" {
@@ -126,13 +129,14 @@ variable "bucket_location" {
 
 variable "auto_create_network" {
   description = "Create the default network"
-  default     = "false"
+  type        = bool
+  default     = false
 }
 
 variable "disable_services_on_destroy" {
   description = "Whether project services will be disabled when the resources are destroyed"
-  default     = "true"
-  type        = string
+  default     = true
+  type        = bool
 }
 
 variable "default_service_account" {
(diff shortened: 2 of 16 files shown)


ingwarr:
@aaron-lane Conflicts resolved I think we need to add a new variable `shared_vpc_subnets_count` to README.md and examples

modules/core_project_factory/main.tf
```diff
@@ -172,8 +172,8 @@ resource "google_project_services" "project_services_authority" {
 /******************************************
   Shared VPC configuration
  *****************************************/
-resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
-  count = var.shared_vpc != "" ? 1 : 0
+resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment"
+  count = var.shared_vpc_subnets_count != null ? 1 : 0
```
kopachevsky: does it make sense to create boolean variable for `var.shared_vpc_subnets_count != null` ?

modules/core_project_factory/main.tf
```diff
@@ -283,7 +283,7 @@ resource "google_service_account_iam_member" "service_account_grant_to_group" {
   Account on shared VPC
  *****************************************************************************************************************/
 resource "google_project_iam_member" "controlling_group_vpc_membership" {
-  count = var.shared_vpc != "" && length(compact(var.shared_vpc_subnets)) == 0 ? local.shared_vpc_users_length : 0
+  count = var.shared_vpc_subnets_count != null && length(compact(var.shared_vpc_subnets)) == 0 ? local.shared_vpc_users_length : 0
```
kopachevsky: same for length(compact(var.shared_vpc_subnets)) , we can define value of this expression it as a local variale and just use it?

modules/core_project_factory/main.tf
ingwarr: Yes it makes sense, I'll try to test it

modules/core_project_factory/main.tf
ingwarr: Better use the boolean variable from your previous comment if it will work

ingwarr:
> I'm not sure this is the correct solution to the issue. The error may be resolved because we've removed `var.shared_vpc` from the counts, but what is the benefit of adding `var.shared_vpc_subnets_count` when the counts already use `length(compact(var.shared_vpc_subnets))`?
> 
> Since the `core_project_factory` submodule is private, we have some flexibility to introduce breaking changes to that submodule without affecting the root module. What if we replace `var.shared_vpc` with a boolean variable in the counts, and set that variable based on the contents of `var.shared_vpc` in the root module? This change would also be necessary in the `gsuite_enabled` submodule.
> 
> Example of `main.tf` in the root module:
> 
> ```diff
> module "project-factory" {
>   source = "./modules/core_project_factory"
> 
>   group_email                 = module.gsuite_group.email
>   group_role                  = var.group_role
>   lien                        = var.lien
>   manage_group                = var.group_name != "" ? "true" : "false"
>   random_project_id           = var.random_project_id
>   org_id                      = var.org_id
>   name                        = var.name
>   project_id                  = var.project_id
>   shared_vpc                  = var.shared_vpc
> + shared_vpc_enabled          = var.shared_vpc != ""
>   billing_account             = var.billing_account
>   folder_id                   = var.folder_id
>   sa_role                     = var.sa_role
>   activate_apis               = var.activate_apis
>   apis_authority              = var.apis_authority
>   usage_bucket_name           = var.usage_bucket_name
>   usage_bucket_prefix         = var.usage_bucket_prefix
>   credentials_path            = var.credentials_path
>   shared_vpc_subnets          = var.shared_vpc_subnets
>   labels                      = var.labels
>   bucket_project              = var.bucket_project
>   bucket_name                 = var.bucket_name
>   bucket_location             = var.bucket_location
>   auto_create_network         = var.auto_create_network
>   disable_services_on_destroy = var.disable_services_on_destroy
>   default_service_account     = var.default_service_account
>   disable_dependent_services  = var.disable_dependent_services
> }
> ```

@aaron-lane  I've implemented this suggestion and it works, wonderful, thank you!!!

ingwarr:
@aaron-lane I think I need to rebase it

aaron-lane:
There are multiple failures within the project-factory-shared-vpc control of the full-local instance: https://concourse.infra.cft.tips/builds/5646#L5d1c47b5:647 (search for "expected" to find them quickly).

ingwarr:
> There are multiple failures within the project-factory-shared-vpc control of the full-local instance: https://concourse.infra.cft.tips/builds/5646#L5d1c47b5:647 (search for "expected" to find them quickly).

@aaron-lane I don't have appropriate privileges to run the full test locally Error: [ERROR] Error creating group: googleapi: Error 403: Not Authorized to access this resource/api, forbidden

        on ../../../modules/gsuite_enabled/main.tf line 46, in resource "gsuite_group" "group":
        46: resource "gsuite_group" "group" {

bohdanyurov-gl:
@ingwarr Issue is probably related to this:
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2d181abb0da9b8dc4b8090ba3504764488ce9a6f/modules/gsuite_enabled/variables.tf#L67
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2d181abb0da9b8dc4b8090ba3504764488ce9a6f/modules/gsuite_enabled/main.tf#L47

Looks like a regression. Please also correctly rebase your changes.

modules/core_project_factory/main.tf
```diff
@@ -443,7 +440,7 @@ resource "google_storage_bucket_iam_member" "api_s_account_storage_admin_on_proj
 resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
   provider = google-beta
 
-  count = local.gke_shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) != 0 ? length(var.shared_vpc_subnets) : 0
+  count = var.shared_vpc_enabled && length(compact(var.shared_vpc_subnets)) != 0 ? length(var.shared_vpc_subnets) : 0
```
aaron-lane: This logic should not be changed. The definition of `local.gke_shared_vpc_enabled` must be updated to use `var.shared_vpc_enabled`.

modules/core_project_factory/main.tf
```diff
@@ -470,8 +467,7 @@ resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
   container.hostServiceAgentUser role granted to GKE service account for GKE on shared VPC
  *****************************************/
 resource "google_project_iam_member" "gke_host_agent" {
-  count = local.gke_shared_vpc_enabled ? 1 : 0
-
+  count = var.shared_vpc_enabled ? 1 : 0
```
aaron-lane: Same issue about `local.gke_shared_vpc_enabled`.

modules/core_project_factory/variables.tf
```diff
@@ -147,3 +147,8 @@ variable "disable_dependent_services" {
   type        = string
 }
 
+variable "shared_vpc_enabled" {
+  description = "Existing of shared vpc"
+  type        = bool
+  default     = false
```
aaron-lane: This should not have a default. We want to ensure that both public modules explicitly pass this argument.

variables.tf
```diff
@@ -148,3 +148,9 @@ variable "disable_dependent_services" {
   type        = string
 }
 
+variable "shared_vpc_enabled" {
```
aaron-lane: This variable should not be defined here.

modules/core_project_factory/variables.tf
ingwarr: Done

variables.tf
ingwarr: Removed

modules/core_project_factory/main.tf
ingwarr: Fixed

modules/core_project_factory/main.tf
ingwarr: Fixed

modules/core_project_factory/main.tf
```diff
@@ -43,7 +43,7 @@ locals {
     google_project.main.number,
   )
   api_s_account_fmt      = format("serviceAccount:%s", local.api_s_account)
-  gke_shared_vpc_enabled = var.shared_vpc != "" && contains(var.activate_apis, "container.googleapis.com") ? "true" : "false"
+  gke_shared_vpc_enabled = var.shared_vpc_enabled && contains(var.activate_apis, "container.googleapis.com") ? true : false
```
bohdanyurov-gl: you don't need ternary operator here

modules/core_project_factory/variables.tf
```diff
@@ -143,7 +147,13 @@ variable "default_service_account" {
...
 variable "disable_dependent_services" {
   description = "Whether services that are enabled and which depend on this service should also be disabled when this service is destroyed."
-  default     = "true"
-  type        = string
+  default     = true
+  type        = bool
+}
+
+variable "shared_vpc_enabled" {
+  description = "Existing of shared vpc"
```
bohdanyurov-gl: "If shared VPC should be used"

variables.tf
```diff
@@ -16,7 +16,8 @@
 
 variable "random_project_id" {
   description = "Enables project random id generation. Mutually exclusive with project_id being non-empty."
-  default     = "false"
+  type = bool
```
bohdanyurov-gl: formatting

variables.tf
```diff
@@ -121,19 +123,20 @@ variable "bucket_location" {
...
-  type        = string
+  default     = false
+  type        = bool
 }
 
 variable "disable_services_on_destroy" {
   description = "Whether project services will be disabled when the resources are destroyed"
-  default     = "true"
-  type        = string
+  default     = true
```
bohdanyurov-gl: Not related to your change, please revert

variables.tf
```diff
@@ -121,19 +123,20 @@ variable "bucket_location" {
...
-  default     = "false"
+  type = bool
+  default     = false
 }
 
 variable "lien" {
   description = "Add a lien on the project to prevent accidental deletion"
-  default     = "false"
-  type        = string
+  default     = false
```
bohdanyurov-gl: Not related to your change, please revert

variables.tf
```diff
@@ -121,19 +123,20 @@ variable "bucket_location" {
 
 variable "auto_create_network" {
   description = "Create the default network"
-  default     = "false"
+  type = bool
```
bohdanyurov-gl: Not related to your change, please revert

variables.tf
```diff
@@ -68,7 +69,8 @@ variable "sa_role" {
 
 variable "apis_authority" {
   description = "Toggles authoritative management of project services."
-  default     = "false"
```
bohdanyurov-gl: Not related to your change, please revert

variables.tf
bohdanyurov-gl: Not related to your change, please revert

variables.tf
```diff
@@ -144,7 +147,12 @@ variable "default_service_account" {
 
 variable "disable_dependent_services" {
   description = "Whether services that are enabled and which depend on this service should also be disabled when this service is destroyed."
-  default     = "true"
-  type        = string
```
bohdanyurov-gl: Not related to your change, please revert

variables.tf
```diff
@@ -144,7 +147,12 @@ variable "default_service_account" {
...
+  default     = true
+  type        = bool
 }
 
+variable "shared_vpc_enabled" {
+  description = "Existing of shared vpc"
```
bohdanyurov-gl: Better description

docs/upgrading_to_project_factory_v1.0.md
```diff
@@ -18,12 +18,12 @@ while the "project-factory-gsuite" module does.
 /// @file main.tf
 
 provider "google" {
-  credentials = "${file(var.credentials_path)}"
```
bohdanyurov-gl: Please remove all changes except shared_vpc configs

bohdanyurov-gl:
@aaron-lane changes that you've proposed wouldn't work (actually, @ingwarr has already verified that they don't). The primary reason is that your variable still depends on computed value, so terraform can not build full graph on plan stage.

I've suggested @ingwarr to implement separate variable "shared_vpc_enabled", that doesn't rely on any computed resources at all.

Please note, that while this is out of context of the issue, I know for sure that the same problem is going to happen with shared_vpc_subnets (they are not covered in tests and issue create don't use them). @ingwarr please open a separate issue to fix subnets as well.

modules/core_project_factory/variables.tf
ingwarr: Done

modules/core_project_factory/main.tf
ingwarr: Fixed

docs/upgrading_to_project_factory_v1.0.md
ingwarr: ok

variables.tf
ingwarr: Reverted

variables.tf
ingwarr: Reverted

variables.tf
ingwarr: reverted

variables.tf
ingwarr: reverted

variables.tf
ingwarr: Done

docs/upgrading_to_project_factory_v1.0.md
ingwarr: Done

variables.tf
ingwarr: Reverted

variables.tf
ingwarr: ok

modules/core_project_factory/main.tf
ingwarr: Fixed

ingwarr:
image
image
image
image
image
image

PR #245, opened 2019-07-10
WIP: Test faster validation method

glarizza:
This PR has been opened to test a faster Terraform validation method that was pulled from the Network module.


glarizza:
Opened this PR to test out validation being run through CI; this is no longer necessary

Issue #243, opened 2019-07-10
Add support for temporarily generated access tokens


rjerrems:
It would be great to support temporarily generated access tokens with this module. Right now, if temporary tokens are used with Terraform provider configuration the module mostly works - except for cases where gcloud commands are used or when the preconditions python script runs as they both assume they are using a static credentials file. An example of my provider setup looks like this:
```
locals {
  tf_sa = "${var.terraform_service_account}"
}

provider "google" {
  alias = "impersonate"

  scopes = [
    "https://www.googleapis.com/auth/cloud-platform",
    "https://www.googleapis.com/auth/userinfo.email",
  ]
}

data "google_service_account_access_token" "default" {
  provider               = "google.impersonate"
  target_service_account = "${local.tf_sa}"
  scopes                 = ["userinfo-email", "cloud-platform"]
  lifetime               = "600s"
}

provider "google" {
  access_token = "${data.google_service_account_access_token.default.access_token}"
  version      = "~> 2.1"
}

provider "google-beta" {
  access_token = "${data.google_service_account_access_token.default.access_token}"
  version      = "~> 2.1"
}
```

If using a setup as mentioned as above, a possible work around in the case of using gcloud commands is to use the impersonation flag and pass the service account to impersonate like this:
```
gcloud compute instances list --impersonate-service-account= "${local.tf_sa}"
``` 
which would work in the example case. If an additional optional parameter for a service account were added, a shell wrapper for gcloud commands could check if that exists and use the impersonation variation of the gcloud command.

Although the caveat with this approach is it makes the assumption that there is a underlying account which has a credential source either using a metadata endpoint from within GCP or a local credential that is able to generate a token for that service account - it would not work in the case where you have used a external tool to generate a temporary token.

I would be happy to write up a PR for this if this would useful for the module.

mikhail-khodorovskiy:
+ 1!
Just spent a day trying to figure out why the default service account can't be deleted while impersonation is used.

morgante:
Fixed by #243.

PR #246, opened 2019-07-16
Adding support for service account impersonation and short lived tokens

rjerrems:
Hi Guys,

Here is a first cut at adding support for short lived tokens / service account impersonation to address this issue:

https://github.com/terraform-google-modules/terraform-google-project-factory/issues/243

I have refactored the service account deletion script (renamed to modify-service-account.sh) and updated the preconditions checking script.

Before merge I will need to work on getting tests to pass but was having some trouble getting the full test suite to run locally.

Thanks


modules/core_project_factory/main.tf
```diff
@@ -169,6 +170,18 @@ resource "google_project_services" "project_services_authority" {
   depends_on = [google_project.main]
 }
 
+resource "google_project_service" "iam_credentials_api" {
```
morgante: Instead of a separate service block, we should inject this into the `activate_apis`. (In particular, this would break in the case of authoritatively setting the project services.)

modules/core_project_factory/main.tf
```diff
@@ -193,64 +206,50 @@ data "null_data_source" "default_service_account" {
...
 
-/******************************************
-  Default compute service account deletion
- *****************************************/
-resource "null_resource" "delete_default_compute_service_account" {
-  count = var.default_service_account == "delete" ? 1 : 0
+/************************************************************************************
+  Default compute service account action. Options are keep, depriviledge, delete.
+ ***********************************************************************************/
+resource "null_resource" "modify_default_compute_service_account" {
```
morgante: Renaming the resource is likely to lead to some issues with existing projects (causing state migrations). I'd prefer if we kept these as two separate null_resources.

modules/core_project_factory/main.tf
```diff
@@ -193,64 +206,50 @@ data "null_data_source" "default_service_account" {
...
+    google_project_service.iam_credentials_api
   ]
 }
 
 /******************************************
   Default Service Account configuration
  *****************************************/
 resource "google_service_account" "default_service_account" {
-  account_id = "project-service-account"
+  account_id   = "project-service-account"
```
morgante: Make sure to run the format/linting commands.

modules/core_project_factory/variables.tf
```diff
@@ -97,6 +98,11 @@ variable "credentials_path" {
   default     = ""
 }
 
+variable "impersonate_service_account" {
+  description = "An optional service account to impersonate. If this service account is not specified, Terraform will fall back to credential file or Application Default Credentials."
```
morgante: Should we note this option conflicts with specifying a `credentials_path` to project factory?

rjerrems: Yep agreed - I will update this

modules/core_project_factory/main.tf
rjerrems: No problem, this formatting change was the result of running `terraform fmt` FYI, running make -s has a bunch of errors locally for me for files I haven't touched, perhaps I am using a different version of shellcheck. Mine is 0.4.7, here is a sample of the errors:

```
In ./helpers/setup-sa.sh line 171:
\ \ - serviceAccount:${SA_ID}\\
^-- SC1117: Backslash is literal in "\ ". Prefer explicit escaping: "\\ ".
  ^-- SC1117: Backslash is literal in "\ ". Prefer explicit escaping: "\\ ".


In ./helpers/setup-sa.sh line 172:
\ \ role: roles/billing.user" policy-tmp-$$.yml && rm policy-tmp-$$.yml.bak
^-- SC1117: Backslash is literal in "\ ". Prefer explicit escaping: "\\ ".
  ^-- SC1117: Backslash is literal in "\ ". Prefer explicit escaping: "\\ ".
```

modules/core_project_factory/main.tf
rjerrems: This has been fixed in the latest commit and is injected in the locals block.

modules/core_project_factory/main.tf
rjerrems: Cool make sense, this has been fixed in the latest commit.

rjerrems:
I have run through the integration tests three times from the last commit and it looks like it is working most of the time, although I have seen this error once (and a couple times on previous commit)

```
       Error: Invalid count argument
       
         on ../../../modules/core_project_factory/main.tf line 113, in resource "null_resource" "check_if_shared_vpc_subnets_contains_items_with_invalid_name":
        113:   count = length(
        114:     compact(null_resource.shared_vpc_subnet_invalid_name.*.triggers.name),
        115:   ) == 0 ? 0 : 1
       
       The "count" value depends on resource attributes that cannot be determined
       resources that the count depends on.
```

Is this a known issue or something I may have introduced?

aaron-lane:
@rjerrems that issue is a regression which exists on the master branch so it's not caused by your changes.

rjerrems:
Ok cool - good to know. Thanks @aaron-lane

rjerrems:
Hi Guys, Just giving this a nudge - let me know if there is anything that needs to be worked on for this?

PR #266, opened 2019-09-13
make project check more efficient

ptone:
When one has access to large numbers of projects, verifying by listing and filtering can be a very slow operation. Since we have the project ID, checking for its existence directly is much faster.


morgante:
Thanks for the contribution, looks like there's a linting issue:

```
In ./helpers/setup-sa.sh line 47:
SEED_PROJECT="$(gcloud projects describe --format="value(projectId)" $2)"
                                                                     ^-- SC2086: Double quote to prevent globbing and word splitting.
```

morgante:
Superseded by https://github.com/terraform-google-modules/terraform-google-project-factory/pull/338

PR #268, opened 2019-09-23
Fabric module: fix custom roles outputs

ludoo:


ludoo:
Build triggers seem to be misconfigured...

PR #269, opened 2019-09-24
add roles/resourcemanager.folderCreator to seed sa [helper]

emalloy:


morgante:
I'm not quite sure why project factory would need folder creator? Creating projects shouldn't require this permission level.

emalloy:
@morgante Correct. This is a misplaced permission addition. Will be following up later and closing for now.

Issue #272, opened 2019-10-01
`project_services` module disables too many APIs on change


ndmckinley:
This issue was found here: https://github.com/terraform-providers/terraform-provider-google/issues/4570#issuecomment-536801048.

When you have a list of services, and you change one of the services, the services after that one in the list will be renumbered.  The renumbering leads to a destroy / create for each service, which could fail if some of the services are dependents of other services which are active.

It seems to me that the right behavior here is to have disable_on_destroy default to false so that modifications can be made safely.  What do you think?

ndmckinley:
Not that it is a big help to change `true` to `false` or anything, but I went and proposed that solution as a PR.  I'll watch to see if there are any tests or anything I should take care of here as well.

morgante:
Thanks! I think we might actually be better off switching to a for loop instead of count.

ndmckinley:
Oh yeah, that's way better.

Chupaka:
But then it will make the module unusable in TF 0.11, I believe. Kind of important for big installations of TFE having trouble to upgrade to 0.12 due to HashiCorp devs made output name restrictions without first documenting it and after those installations were deployed :)

morgante:
Well, we no longer provide official support for <0.12 on the module anyways so updates can use 0.12 functionality.

The old version of the module which can be used with 0.11 is [v.2.4.1](https://registry.terraform.io/modules/terraform-google-modules/project-factory/google/2.4.1/submodules/project_services).

ndmckinley:
Hi there,

Not sure I'm the right assignee for this, since I think the solution has actually already been merged.

PR #273, opened 2019-10-01
`disable_services_on_destroy` nondestructive by default

ndmckinley:
This would address #272.


morgante:
As discussed, would prefer a for loop approach.

PR #284, opened 2019-10-15
Use project_services module and convert to for_each

### LINKED ISSUE #282: Refactor service activation
We need to improve service management in this module to address issues with duplication and potential race conditions.

We should:
1. Remove [authoritative mode](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L165), as described in #279 
2. Update the [core_project_factory](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L153) to use the [project_services module](https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/project_services)
3. Update the project_services module to use `for_each` instead of `count`.
4. Provide a migration guide + script for Project Factory users to upgrade to the new version without any Terraform plan diff appearing. (cf. [this guide](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/docs/upgrading_to_project_factory_v1.0.md#upgrading-to-project-factory-v10))

Rycieos:
Convert the project_services module to use for_each instead of count,
and use that module in the core project factory.

Remove the var.apis_authority option.

Add to helpers/migrate.py to also convert this project services
migration. The old GSuite migration will still work. To make this
migration work, it needs to read a value from a resource, which means
calling terraform state show, which is slow.

This is heavily blocked by hashicorp/terraform#22301, as the migrate
script fails to state move the instances of count to instances of
for_each without the for_each resource existing first. As it is, the
migrate script fails with this error:
```
Cannot move to
module.my-project.module.project-factory.module.project_services.google_project_service.project_services["cloudbilling.googleapis.com"]:
module.my-project.module.project-factory.module.project_services.google_project_service.project_services
does not exist in the current state.
```

This is also partially blocked by hashicorp/terraform#10462. As this
moved the project services into a module, the depends_on in the
project_service resource block can't be used in the module block because
of this issue. Since the project_id parameter is referencing the
google_project, it might still work correctly.

Resolves #282

### DIFF
--- a/docs/upgrading_to_project_factory_v4.0.md
+++ b/docs/upgrading_to_project_factory_v4.0.md
@@ -1,0 +1,208 @@
+# Upgrading to Project Factory v4.0
+
+The v4.0 release of Project Factory is a backwards incompatible release and
+features minor changes, specifically with how Google Project Services (APIs)
+are managed. A state migration script is provided to update your existing states
+to minimize or eliminate resource re-creations.
+
+Note that upgrading requires you to have Python 3.7 installed.
+
+## Migration Instructions
+
+This migration was performed with the following example configuration.
+
+```hcl
+/// @file main.tf
+
+provider "google" {
+  credentials = "${file(var.credentials_path)}"
+}
+
+module "my-project" {
+  source            = "terraform-google-modules/project-factory/google"
+  version           = "v3.0.0"
+  random_project_id = "true"
+  name              = "pf-project-services-migrate"
+  org_id            = "${var.org_id}"
+  billing_account   = "${var.billing_account}"
+  credentials_path  = "${var.credentials_path}"
+  activate_apis = [
+    "compute.googleapis.com",
+    "iamcredentials.googleapis.com",
+  ]
+}
+```
+
+### Update the project-factory source
+
+Update the project-factory module source to the Project Factory v4.0.0 release.
+
+```diff
+diff --git i/main.tf w/main.tf
+index d876954..ebb3b1e 100755
+--- i/main.tf
++++ w/main.tf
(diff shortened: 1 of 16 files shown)


Rycieos:
Updated with requested changes.

Note that after the migrate each resource instance will show `disable_dependent_services` and `disable_on_destroy` need to be set to `true`, because as a Terraform state only value, it can't be gotten from importing. This makes this `tf state rm; tf import` workaround functional but messy.

docs/upgrading_to_project_factory_v4.0.md
```diff
@@ -0,0 +1,208 @@
...
+```
+
+### Download the state migration script
+
+ ```
+curl -O https://raw.githubusercontent.com/terraform-google-modules/terraform-google-project-factory/v4.0.0/helpers/migrate4.py
+chmod +x migrate4.py
+```
+
+### Reinitialize Terraform
```
morgante: FYI I don't think the state download is needed for Terraform 0.12. Can probably skip straight to running the script.

modules/project_services/outputs.tf
```diff
@@ -16,9 +16,6 @@
 
 output "project_id" {
   description = "The GCP project you want to enable APIs on"
-  value = element(
-    concat(google_project_service.project_services.*.project, [""]),
-    0,
-  )
+  value       = var.enable_apis && length(var.activate_apis) > 0 ? var.project_id : ""
```
morgante: We still need this to pass through the project_id from the `google_project_service` resource (to preserve the dependency graph), not directly from var.project_id.

PR #290, opened 2019-10-17
Added pip3_extra_flags variable to allow --user param to be passed to…

ideasculptor:
… `pip3 install`

I figured I'd send a draft PR with just the code for setting the extra pip3 param, to make sure you were OK with the mechanism by which I am setting it. It is tested to work correctly.


morgante:
Looks like a good start.

ideasculptor:
Could you let me know what is failing here? It passes when I run the tests locally.

modules/core_project_factory/scripts/preconditions.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
ideasculptor: I'm not sure how folks will feel about this, but I could not figure out a way to extract a parameter from the params destined for the python script without relying on external dependencies (sed, grep) unless it runs in bash, which allows arrays.

modules/core_project_factory/scripts/preconditions.sh
```diff
@@ -14,17 +14,50 @@
...
+}
+
 if command -v python3 1>/dev/null; then
     BASEDIR="$(dirname "$0")"
 
+    extract_pip_flags "$@"
+
     if command -v pip3 1>/dev/null; then
-        exec "pip3" "install" "-r" "$BASEDIR/preconditions/requirements.txt"
+        if [ "${#PIP_FLAGS[*]}" == "0" ]; then
```
ideasculptor: Technically, this conditional isn't actually required.  Because the linter wants all variables to be quoted, and sending an empty string as one of the params to exec does cause the empty param to be received by pip, I can't make it work correctly and pass lint without this extra conditional.

Technically, this is much more correct:

`exec "pip3" "install" ${PIP_FLAGS[*]} "-r" "path/to/requirements.txt"`

Since each element of the array will be sent separately.  But the linter wants to force quotes around the PIP_FLAGS reference.  If I put it in quotes, then pip objects to an empty arg being received.  So I added a conditional to prevent it from being sent at all unless it actually has a value in it, otherwise it is included, within a quoted string.

morgante: We could also simply disable the linter check on this line if you think that'd improve the readability. Really, this seems a bit overcomplicated for simply injecting an extra flag—I'm surprised we need an `extract_pip_flags` flag at all.

One possible alternative/simpler implementation would be to only expose a "install_preconditions_as_user" variable which adds the required flag.

ideasculptor: Sure, but the difficulty is in how to actually pass that information to the preconditions.sh script.  If it goes in the args, then you have to get it back out before they are passed to the python script.  Otherwise, you need either a positional param we can shift off the front, or 2 versions of the script.  I'm happy to implement it either way, if someone wants to express a preference.  

I agree that this is a complex way to solve the problem, but it does keep duplicated code to a minimum and results in minimum impact to the user of the module while delivering the flexibility to pass arbitrary params to pip, rather than only allowing them to select between 2 options.  It is easy to imagine a requirement for a fair number of available options, especially for enterprise users, who can be stuck behind firewalls and internal artifact repositories, etc.

These are just the params that stand out as likely candidates for at least some users of the module.

```
  --user                      Install to the Python user install directory for your platform. Typically ~/.local/, or %APPDATA%\Python on Windows. (See the Python documentation for site.USER_BASE for full details.)
  --proxy <proxy>             Specify a proxy in the form [user:passwd@]proxy.server:port.
  --trusted-host <hostname>   Mark this host as trusted, even though it does not have valid or any HTTPS.
  --cert <path>               Path to alternate CA bundle.
  --client-cert <path>        Path to SSL client certificate, a single file containing the private key and the certificate in PEM format.
```

I did realize the PR as it stands right now, has a bug in it which will actually break if you set pip3_extra_flags to be '--user', anyway, since it will recognize that as a flag intended for the python script instead of pulling it into the pip flags.  I know how to fix it, but I'll wait until I hear a solution preference.

morgante: What if we simple separated this into two different `local-exec` scripts? One could handle installing the pip dependencies, while the second actually executes the Python script.

morgante:
@ideasculptor Thoughts on the suggested approach?

ideasculptor:
To be honest, I mostly concluded that the lazy approach - manually pip installing the requirements.txt file after erroring out - was far easier relative to the effort required to do it correctly (activate virtual env, install into there, run script, then deactivate virtual env). Using a virtual env is certainly a ‘more correct’ solution than installing globally, whether the host OS allows it without —user or not.  Unfortunately, that comes with its own set of complexities, since different python versions require different commands to set up a virtual env. 

I suspect documenting the package requirements up front and letting users handle it has a lot less short term development friction, though the virtual env solution would likely prove useful throughout the module library, not just here.

morgante:
@ideasculptor Got it, I think I agree it is likely simpler to install requirements up-front. I think the main place this becomes an issue is in CI environments/Terraform Enterprise where they need to be installed within a container.

ideasculptor:
I agree, but even then it is pretty easy to install them when building the container so long as they are documented up front.

Then you could just eliminate the step which installs requirements entirely, in order to avoid conflicts and complexity. I suspect that would simplify the provisioners, only at the expense of a fairly reasonable up-front dependency resolution step.

morgante:
Agreed.

PR #295, opened 2019-10-26
Script Helper Update

bg-master:
Update the helper script to grant the necessary roles to the Seed Service. Add optional roles/resourcemanager.folderViewer on the folder where the project will be created.

### DIFF
--- a/helpers/setup-sa.sh
+++ b/helpers/setup-sa.sh
@@ -137,6 +137,14 @@ gcloud projects add-iam-policy-binding \
   --role="roles/resourcemanager.projectIamAdmin" \
   --user-output-enabled false
 
+# Grant roles/storage.admin to the Seed Service Account on the Seed Project
+echo "Adding role roles/storage.admin..."
+gcloud projects add-iam-policy-binding \
+  "${SEED_PROJECT}" \
+  --member="serviceAccount:${SA_ID}" \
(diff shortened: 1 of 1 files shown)


helpers/setup-sa.sh
```diff
@@ -17,73 +17,138 @@
...
+  --project "${SEED_PROJECT}" create ${SA_NAME} \
+  --display-name ${SA_NAME}
 
 echo "Downloading key to credentials.json..."
 gcloud iam service-accounts keys create "${KEY_FILE}" \
-    --iam-account "${SA_ID}" \
+  --iam-account "${SA_ID}" \
+  --user-output-enabled false
+
+if [[ $FOLDER_ID == "" ]]; then
```
aaron-lane: This condition will never trigger based on the earlier exit [here](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/295/files#diff-f3fdb2558703bcb409cbf6fdc6d5b5ecR114-R117).

bg-master: The first check occurs only if the "folder" parameter is passed, and exit if the project folder does not exist on gcp.
If the "folder" parameter is not passed, the earlier check is always skipped, and this check is necessary in order not to apply the permissions related to the folder.

PR #302, opened 2019-10-29
use count instead of for_each to support conditional enabling

milesmatthias:
Addresses issues discovered in https://github.com/terraform-google-modules/terraform-google-memorystore/pull/15

### DIFF
--- a/modules/project_services/main.tf
+++ b/modules/project_services/main.tf
@@ -18,9 +18,9 @@
   APIs configuration
  *****************************************/
 resource "google_project_service" "project_services" {
-  for_each                   = toset([for api in var.activate_apis : api if var.enable_apis])
+  count                      = var.enable_apis == "true" ? length(var.activate_apis) : 0
   project                    = var.project_id
-  service                    = each.value
+  service                    = var.activate_apis[count.index]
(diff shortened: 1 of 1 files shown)


morgante:
We specifically didn't do this because count causes issues with reordering APIs.

Could we combine count *and* for_each?

milesmatthias:
No, TF doesn't allow combining `count` and `for_each`: https://www.terraform.io/docs/configuration/resources.html#for_each-multiple-resource-instances-defined-by-a-map-or-set-of-strings

milesmatthias:
The other option here is to instruct users to use this module conditionally by passing in an empty list, something like this:

```
  activate_apis = var.enable_apis == "true" ? [
    "redis.googleapis.com",
  ] : []
```

morgante:
Could we create a local which simply does that, then for_each over the local?

milesmatthias:
I tried that approach, and it generates the same error.

morgante:
Ok I'm going to try some possible solutions myself.

morgante:
@milesmatthias PTAL at #303 and confirm it works for your module.

Issue #306, opened 2019-11-01
modules only allow single shared_vpc


ideasculptor:
I have a service-project which will need to use the network of several host projects - one for each environment that the service-project resources will run within.

All of the project-factory modules assume just a single host vpc, and only set up the iam resources, especially the kubernetes-specific iam resources, for a single host project.

If, as I did, you then assume that we should not use the shared_vpc functionality exposed in these modules, but should instead use the network modules, which include a module specifically for enabling host network access in a service project, you are in for a rude awakening, because that module does not do much of the logic that is contained in the project-factory modules.  Even worse, it uses the authoritative iam_binding, so if you use that module on a project which was correctly set up in the project-factory, you will break things when the iam_binding gets overridden without all of the correct values.  Meanwhile, the project-factory modules don't allow you to specify per-subnet access controls.  

The end result is that no matter which strategy you use to set up network access, you are pretty much guaranteed to run into problems if you do anything more sophisticated than the examples - single host project, single client project, all subnets, with subnets created prior to attaching the client to the host.

morgante:
GCP doesn't support associating a service project with multiple host projects: "each service project can only be attached to a single host project" from https://cloud.google.com/vpc/docs/shared-vpc

PR #307, opened 2019-11-01
Enable multiple host networks

ideasculptor:
The project factory modules can only attach to a single host project at create-time, but it is not uncommon for a service project to be associated with more than one host project.  This PR breaks out the resources that set up VPC sharing into their own module, so that it can be called repeatedly in order to attach multiple host projects.  The project-factory modules, themselves, have unchanged behaviour, but a caller now has flexibility to add several more host projects (and subnets) to a service project after the project factory module returns.  This new module probably deprecates the terraform-google-modules/terraform-google-network/google//modules/fabric-net-svpc-access module.

fixes https://github.com/terraform-google-modules/terraform-google-project-factory/issues/306


ideasculptor:
Details about that test failure would be handy. If there's a way you can give me read-only access to cloud-build trigger output, that would sure be handy.  I'd be happy to provide the address of a gsuite group or you can use my email address, which you already have. Running them locally just to see the output is both time consuming and not necessarily effective, depending on the failure.

ideasculptor:
The error I'm seeing is as follows - it seems unlikely to be related to the PR, since it isn't even getting to the point where the created project can have anything applied to it. I've run it several times with the same result.

```
       module.project-factory.module.project-factory.google_project.main: Still creating... [9m59s elapsed]
       module.project-factory.module.project-factory.google_project.main: Still creating... [10m9s elapsed]
       
       Error: Error enabling the Compute Engine API required to delete the default network: Error waiting for Enable Project "pf-ci-test-minimal-ffc7ac-a696" Services: [compute.googleapis.com]: timeout while waiting for state to become 'done: true' (last state: 'done: false', timeout: 10m0s)
       
         on ../../../modules/core_project_factory/main.tf line 114, in resource "google_project" "main":
        114: resource "google_project" "main" {
       
```

morgante:
GCP doesn't support associating a service project with multiple host projects: "each service project can only be attached to a single host project" from https://cloud.google.com/vpc/docs/shared-vpc

ideasculptor:
Well, that's an annoying oversight.  It was a gcloud sales engineer who suggested I structure things with a 'shared service project' (his words, IIRC) in the first place, a couple of months ago, so it never even occurred to me to verify that it was supported.  I just assumed...  Oh well.

morgante:
I'd assume he meant a shared host project (where you can host shared services).

PR #314, opened 2019-11-14
Add option to disable default compute SA

### LINKED ISSUE #313: Deletion of Default Compute Service Account prevents all VM usage
The default for the default compute service account is to delete it on project creation. This will cause many problems if the Compute API is used, preventing any VMs from being created. Obviously, we could change our setting to "keep" instead, but we have had many teams not realize that the default is delete and not make that change. Since the service account can only be undeleted within 30 days, many of my teams have missed that window, tried to create VMs, and realized that their service account is long gone. We have no choice at that point except to create a new project. We tested disabling the Compute API and re-enabling, but the service accounts don't come back. We also contacted Google support, and they affirmed that the only fix is to make a new project.

For this reason, I would prefer that the default is "keep" instead of "delete", since it's less destructive.

If that is too much of a change, I'd like to see another option added, "disable", which disables the service account. This is a better option than "delete", since it has the same outcome (the account doesn't work), but it is possible to get the account back, unlike which delete which can be unrecoverable. This then should be the default, since it achieves the same outcome, but is completely and easily reversible for those that realize they don't want the default later on.

Related to this, the service account doesn't even exist until the Compute API is enabled. I have tested this with some projects. This is the default:
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/fa21b2296ddebc725f7a3b4f83ca03f713be483e/variables.tf#L80-L84
There are other APIs that are not enabled by default like this, but are instead enabled if a related feature is turned on:
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/fa21b2296ddebc725f7a3b4f83ca03f713be483e/modules/core_project_factory/main.tf#L45
In this case, if the service account impersonation feature is enabled, then the `iamcredentials` API is also enabled.

We could do something similar for the Compute API: if the `shared_vpc_subnets` or `shared_vpc` inputs are specified, then enable the Compute API. This would prevent the service account being created until someone actually wants to use it.

Related: #181, #186

Thoughts?

Rycieos:
Disabling instead of deleting the default service account achieves the
same result: the service account no longer works. But disabling is
possible to undo at any time, where after 30 days a deleted account is
unrecoverable.

I have tested this with creating a new project, and it works as expected.

Thoughts about making this the default instead of delete?

Resolves #313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@ Extending the adopted spec, each change should have a link to its corresponding 
 
 ## [Unreleased]
 
+### Added
+
+- Option to disable the default compute service account. [#313]
+
 ### Fixed
 
 - Fixed an issue with passing an empty list to activate_apis. [#300]
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ determining that location is as follows:
 | bucket\_name | A name for a GCS bucket to create (in the bucket_project project), useful for Terraform state (optional) | string | `""` | no |
 | bucket\_project | A project to create a GCS bucket (bucket_name) in, useful for Terraform state (optional) | string | `""` | no |
 | credentials\_path | Path to a service account credentials file with rights to run the Project Factory. If this file is absent Terraform will fall back to Application Default Credentials. | string | `""` | no |
-| default\_service\_account | Project default service account setting: can be one of `delete`, `depriviledge`, or `keep`. | string | `"delete"` | no |
+| default\_service\_account | Project default service account setting: can be one of `delete`, `depriviledge`, `disable`, or `keep`. | string | `"delete"` | no |
 | disable\_dependent\_services | Whether services that are enabled and which depend on this service should also be disabled when this service is destroyed. | bool | `"true"` | no |
 | disable\_services\_on\_destroy | Whether project services will be disabled when the resources are destroyed | string | `"true"` | no |
 | domain | The domain name (optional). | string | `""` | no |
@@ -163,7 +163,7 @@ determining that location is as follows:
 
 ### Software
 
--   [gcloud sdk](https://cloud.google.com/sdk/install) >= 206.0.0
+-   [gcloud sdk](https://cloud.google.com/sdk/install) >= 269.0.0
 -   [jq](https://stedolan.github.io/jq/) >= 1.6
 -   [Terraform](https://www.terraform.io/downloads.html) >= 0.12.6
 -   [terraform-provider-google] plugin 2.1.x
--- a/examples/simple_project/README.md
+++ b/examples/simple_project/README.md
@@ -15,7 +15,7 @@ Expected variables:
 |------|-------------|:----:|:-----:|:-----:|
 | billing\_account | The ID of the billing account to associate this project with | string | n/a | yes |
 | credentials\_path | Path to a service account credentials file with rights to run the Project Factory. If this file is absent Terraform will fall back to Application Default Credentials. | string | `""` | no |
-| default\_service\_account | Project default service account setting: can be one of `delete`, `depriviledge`, or `keep`. | string | n/a | yes |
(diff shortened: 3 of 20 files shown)


Rycieos:
I'll get working on the tests.

In regards to the breaking change, I was thinking it wouldn't be. Changing the default would obviously only apply when the user didn't override the default. Which means that their service account would already be deleted (since that is the default), and then disabling would be a no-op, since it wouldn't find any account to delete.

morgante:
> In regards to the breaking change, I was thinking it wouldn't be. Changing the default would obviously only apply when the user didn't override the default. Which means that their service account would already be deleted (since that is the default), and then disabling would be a no-op, since it wouldn't find any account to delete.

Good point, but I think it'd still be a change in the functional behavior of the module using defaults so I think it's likely a breaking change anyways. (Ex. imagine if someone adds another project using the same config, they'd be surprised that one has the SA disabled while the other has it deleted).

Still, more than happy to consider cutting a major release for this.

Rycieos:
Ran into a snag: this disabling functionality only went GA two weeks ago in [version 269.0.0](https://cloud.google.com/sdk/docs/release-notes#26900_2019-10-29).

Not only would this change this project's supported versions of `gcloud`, but also the image used to run the tests has version `255.0.0`. I also checked newer versions of the image, but they have the same `gcloud` version, which seems to be locked based on [this file](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/d6859158c957b2a4fb482933da9d184e6bce665c/infra/build/Makefile#L17)

It seems I have two options:
1. Add `alpha` to the command call, which should just work since the functionality hasn't changed at all.
2. Update the required version of `gcloud` in both the README and the build image.

I think the first option is better, but the second option brings up an interesting point; that the minimum required version in the README is different than the version used to run the tests. Should that be changed?

morgante:
I'd actually lean towards option #2 (Update the required version of gcloud in both the README and the build image.), as it should be compatible in the future.

Rycieos:
Do you want me to make the change to the CFT? I'm not familiar with that project, so I'm not sure I would know what to do.

I'll get working on testing the test changes with a newer version of `gcloud`.

morgante:
> Do you want me to make the change to the CFT? I'm not familiar with that project, so I'm not sure I would know what to do.

If you're able to, it would be helpful! A PR updating the `CLOUD_SDK_VERSION` and `DOCKER_TAG_VERSION_DEVELOPER_TOOLS` in the Makefile should be sufficient.

Rycieos:
Wrapping up with the tests. Do you want me to also bump the required version of `gcloud` in the README? What version, 271 or 269?

morgante:
269

Rycieos:
Tests added. I also added a test to the `full` test suite to cover the default compute service account being deleted. But it doesn't look like that test suite is run? Not sure if I am missing something.

My Ruby isn't the greatest, so feel free to fix that.

morgante:
Yes I think we disable it in CI due to G Suite issues. Can you also update the image version used for CI? https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/build/int.cloudbuild.yaml#L43

Rycieos:
Thanks for pointing that out, I tried to figure out where CI would look for it (figured it wouldn't use the Makefile) but I didn't look hard enough. Let's hope I didn't forget anything else.

Rycieos:
@morgante any help with why the build failed? I can't view the build logs. `make docker_test_integration` runs successfully for me.

morgante:
@Rycieos The error looks spurious, I'm attempting to rerun.

morgante:
@Rycieos It looks like the upgrade to the new devtools image was missing some required changes, I'm handling in #318.

Rycieos:
I've rebased after your PR, so it builds now.

Do you want a separate PR for making `"disable"` the default?

morgante:
> Do you want a separate PR for making "disable" the default?

Yes please!

PR #315, opened 2019-11-15
Hard-code admin email, domain in build

aaron-lane:
This change allows the trigger to have a consistent interface with other modules, in support of https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/414.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@
 # Make will use bash instead of sh
 SHELL := /usr/bin/env bash
 
-DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 0.1.0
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@
 # Make will use bash instead of sh
 SHELL := /usr/bin/env bash
 
-DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 0.1.0
+DOCKER_TAG_VERSION_DEVELOPER_TOOLS := 0
```
morgante: We should talk about this, but I'm against using a global version which can inexplicably break.

morgante:
Incorporated into #318

PR #318, opened 2019-11-18
Upgrade to CFT devtools 0.6.0

morgante:
Removes the necessity of an explicit `make_source.sh` file.


morgante:
> Can you update the tag to 0 to reduce the need for future changes?

I don't really agree with using a generic tag considering new versions can introduce breaking changes.

PR #319, opened 2019-11-19
Change default for default compute SA to disable

Rycieos:
This is a better option than "delete", since it has the same outcome
(the account doesn't work), but it is possible to get the account back,
unlike which delete which can be unrecoverable. This then should be the
default, since it achieves the same outcome, but is completely and
easily reversible for those that realize they don't want the default
later on.

Related #313, #314

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ Extending the adopted spec, each change should have a link to its corresponding 
 
 - Option to disable the default compute service account. [#313]
 
+### Changed
+
+- **Breaking**: Default for default compute service account changed to disable from delete. [#313]
+
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


Rycieos:
Bump.
Anything still needed on this? We are looking forward to this in a new version (as well as the unreleased empty activate_apis fix).

morgante:
@Rycieos We just need a (simple) upgrade guide, like [this one](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/docs/upgrading_to_project_factory_v2.0.md#app-engine-argument-changes) explaining what to change if you want to use the old default.

Rycieos:
I can do that. Gonna be a really short document:
```
The default of default_service_account changed from "delete" to "disable". If you want to keep deleting for new projects, you will need to specify that:

default_service_account = "delete"
```

Issue #216, opened 2019-05-28
Make sure tests do not depend on organization viewer permission


morgante:

morgante:
To ensure that #20 doesn't get repeated.

paulpalamarchuk:
I've made some test, and discovered that core module depends on org-level permissions:
```
       Error: Error reading Organization Not Found : 684124036889: googleapi: Error 403: The caller does not have permission, forbidden

         on ../../../modules/gsuite_group/main.tf line 26, in data "google_organization" "org":
         26: data "google_organization" "org" {
```
That happens because of calling [`gsuite_group` module](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/main.tf#L21) from core-module. That has [data org resource](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/gsuite_group/main.tf#L18)

imrannayer:
Fixed in #320

PR #320, opened 2019-11-21
Fix organization permission requirement

### LINKED ISSUE #20: Organization viewer should not be needed

paulpalamarchuk:
* Fix #20 (#216)


imrannayer:
Fixed in #320

PR #326, opened 2019-12-02
Modify app engine example, fixture, obj for healthchecks

### LINKED ISSUE #227: Add test suite for app_engine submodule
This test suite should leverage the `app_engine` example module.

### LINKED ISSUE #324: Update app_engine.feature_settings type and default
`modules/app_engine/variables.tf` currently defines `feature_settings` as a generic `list` with a default of `[{ enabled = true }]`. The type should be updated to `list(object{ split_health_checks = bool })` since `feature_settings` can only be defined once on `google_app_engine_application` and `split_health_checks` is the only valid argument. The default should be updated to `{ split_health_checks = true}` since `false` enables legacy health checks.

https://www.terraform.io/docs/providers/google/r/app_engine_application.html#feature_settings

bharathkkb:
fixes #324 
fixes #227

### DIFF
--- a/examples/app_engine/main.tf
+++ b/examples/app_engine/main.tf
@@ -14,18 +14,42 @@
  * limitations under the License.
  */
 
-/******************************************
-  Provider configuration
- *****************************************/
 provider "google" {
   version = "~> 2.18.1"
 }
 
+provider "google-beta" {
+  version = "~> 2.18.1"
+}
+
+provider "null" {
+  version = "~> 2.1"
+}
+
+provider "random" {
+  version = "~> 2.2"
+}
+
+resource "random_string" "suffix" {
+  length  = 4
+  special = false
+  upper   = false
+}
+
+module "app-engine-project" {
+  source            = "../../"
+  name              = "appeng-${random_string.suffix.result}"
+  random_project_id = "true"
+  org_id            = var.org_id
+  folder_id         = var.folder_id
+  billing_account   = var.billing_account
+  activate_apis = [
+    "appengine.googleapis.com",
+  ]
+}
+
 module "app-engine" {
-  source           = "../../modules/app_engine"
-  location_id      = var.location_id
-  auth_domain      = var.auth_domain
-  serving_status   = var.serving_status
-  feature_settings = [{ enabled = true }]
-  project_id       = "example-project"
+  source      = "../../modules/app_engine"
+  project_id  = module.app-engine-project.project_id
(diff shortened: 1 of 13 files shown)


examples/app_engine/main.tf
```diff
@@ -14,23 +14,34 @@
...
+resource "random_string" "suffix" {
+  length  = 4
+  special = false
+  upper   = false
+}
+
+module "app-engine-project" {
+  source            = "../../"
+  name              = "appeng-${random_string.suffix.result}"
+  random_project_id = "true"
```
aaron-lane: Booleans no longer require quotation marks.

```suggestion
  random_project_id = true
```

examples/app_engine/main.tf
```diff
@@ -14,23 +14,34 @@
...
+  billing_account   = var.billing_account
+  activate_apis = [
+    "appengine.googleapis.com",
+  ]
 }
 
 module "app-engine" {
-  source           = "../../modules/app_engine"
-  location_id      = var.location_id
-  auth_domain      = var.auth_domain
```
aaron-lane: The `auth_domain` variable should be removed since it is not used.

examples/app_engine/main.tf
```diff
@@ -14,23 +14,34 @@
...
-  serving_status   = var.serving_status
-  feature_settings = [{ enabled = true }]
-  project_id       = "example-project"
+  source      = "../../modules/app_engine"
+  project_id  = module.app-engine-project.project_id
+  location_id = var.location_id
```
aaron-lane: Let's drop `location_id` since we're using the default value.

examples/app_engine/main.tf
bharathkkb: Done

examples/app_engine/main.tf
bharathkkb: I havent used ```auth_domain``` in the tests since I did not know what example domain to use but it is still a valid input to the resource.

examples/app_engine/main.tf
bharathkkb: Done

examples/app_engine/main.tf
aaron-lane: Sorry, I just mean from the example. We aren't using it any more so it should be removed.

aaron-lane: Oh. On a second look, it has already been removed. Ignore me. :sweat_smile:

PR #331, opened 2019-12-11
Adds check for gcloud executable in PATH

marshallford:
I ran into an unclear error message on apply when using this module in a TFE environment where `gcloud` had not be installed. I'm thinking a precondition error message would be helpful, especially for folks running TFE or using Terraform Cloud.

A larger conversation: Can `modify-service-account.sh` be rewritten to not rely on the `gcloud` cli? Perhaps a native module can make changes to the project SA?

Thanks.


marshallford:
@aaron-lane See #336. Thanks!

PR #340, opened 2019-12-16
Clarify docs for project_id and random_project_id

### LINKED ISSUE #337: Confusing descriptions on project_id and random_project_id
The descriptions in the documentation for `project_id` and `random_project_id` are confusing at best, and are probably incorrect. For how I understand what "mutually exclusive" means, I should specify either `project_id=somevalue` or `random_project_id=true`, but not both. 

However, in looking at the code, I see no such restrictions. I can happily specify both, and end up with the project ID being `somevalue-53`, or if I specify the `name=somename`, the project ID will become either `somename-53` or just `somename`, depending on how `random_project_id` is configured. 

Am I missing something in my understanding? Could we remove those "mutually exclusive" statements and perhaps state that the `project_id` will default to the `name` provided? That seems to be how the code is written.

bbrouwer:
`project_id` and `random_project_id` are not mutually exclusive.
Clarify that `project_id` will default to the value in `name`.

fixes #337


modules/core_project_factory/variables.tf
```diff
@@ -39,13 +39,13 @@ variable "manage_group" {
...
 variable "project_id" {
-  description = "If provided, the project uses the given project ID. Mutually exclusive with random_project_id being true."
+  description = "The ID to give the project. If not provided, the `name` will be used."
   type        = string
   default     = ""
 }
 
 variable "random_project_id" {
-  description = "Enables project random id generation. Mutually exclusive with project_id being non-empty."
+  description = "Adds a suffix of 2 random characters to the `project_id`"
```
aaron-lane: This description needs to be corrected. The suffix generator uses 2 bytes, which translates to 4 characters. Please update all instances of this description.

```suggestion
  description = "Adds a suffix of 4 random characters to the `project_id`."
```

bbrouwer: I have applied the requested changes.

bbrouwer:
Is there anything further I need to do to get this merged?

morgante:
Thanks for the contribution!

PR #354, opened 2020-01-06
Feature/330 budget management

### LINKED ISSUE #330: Budget management
Terraform now has a resource for [managing budgets](https://www.terraform.io/docs/providers/google/r/billing_budget.html). We should add support in Project Factory.

# Budget Submodule
We should create a budget submodule with the following inputs:
1. `billing_account`
2. `projects`
3. `display name` (default to `Budget for $projects[0]`)
4. `credit_types_treatment` (default to `INCLUDE_ALL_CREDITS`)
5. `services` (default to null)
6. `amount` (number)
7. `alert_spent_percents`: list of percentages to alert at (compose threshold_rules form this); default to 0.5, 0.7, 1.0
8. `alert_pubsub_topic`: pubsub topic to send alerts to, if not provided do not add the `all_updates_rule` block
9. `create_budget` top-level boolean that determines whether budget is created at all (default true)

# Existing Module
We should add a reference to the `budget` submodule in the main, gsuite_enabled, and shared_vpc modules. It should pass through the billing-account and project from the module.

Additionally, these variables should be exposed at the top level:
1. `budget_amount` (number) - default to null, if set then pass through and set `create_budget` on submodule
2. `budget_pubsub_alert_topic`: pubsub topic to send alerts to, if not provided do not add the `all_updates_rule` block
3. `budget_alert_spent_percents`: list of percentages to alert at (compose threshold_rules form this); default to 0.5, 0.7, 1.0

taylorludwig:
Fixes #330 

New `budget` module is created.
 
If `budget_amount` is set in the main, core, gsuite, of local vpc modules then the budget is also created

Increases google provider to `>= 3.1.0` - See note below about creating a release prior to this so people can target 2.x versions 

Tests created

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ determining that location is as follows:
 | bucket\_location | The location for a GCS bucket to create (optional) | string | `"US"` | no |
 | bucket\_name | A name for a GCS bucket to create (in the bucket_project project), useful for Terraform state (optional) | string | `""` | no |
 | bucket\_project | A project to create a GCS bucket (bucket_name) in, useful for Terraform state (optional) | string | `""` | no |
+| budget\_alert\_pubsub\_topic | The name of the Cloud Pub/Sub topic where budget related messages will be published, in the form of `projects/{project_id}/topics/{topic_id}` | string | `"null"` | no |
+| budget\_alert\_spent\_percents | A list of percentages of the budget to alert on when threshold is exceeded | list(number) | `<list>` | no |
+| budget\_amount | The amount to use for a budget alert | string | `"null"` | no |
 | credentials\_path | Path to a service account credentials file with rights to run the Project Factory. If this file is absent Terraform will fall back to Application Default Credentials. | string | `""` | no |
 | default\_service\_account | Project default service account setting: can be one of `delete`, `deprivilege`, `disable`, or `keep`. | string | `"disable"` | no |
 | disable\_dependent\_services | Whether services that are enabled and which depend on this service should also be disabled when this service is destroyed. | bool | `"true"` | no |
@@ -148,6 +151,7 @@ determining that location is as follows:
 
 | Name | Description |
 |------|-------------|
+| budget\_name | The name of the budget if created |
 | domain | The organization's domain |
 | group\_email | The email of the G Suite group with group_name |
 | project\_bucket\_self\_link | Project's bucket selfLink |
--- a/README.md
+++ b/README.md
@@ -182,7 +182,7 @@ determining that location is as follows:
 
 Starting with version `6.3.0` of this module, `google_billing_budget` resources can now be created. This increases the minimum `terraform-provider-google` version to `3.1.0`
 
-To continue to use a version `>= 2.1, < 3.1` of the google provider pin this module to `6.2.1`
+To continue to use a version `>= 2.1, < 3.1` of the google provider pin this module to `6.2.1`. Or use the `core_project_factory` submodule directly.
 
 ```hcl
 module "project-factory" {
--- a/README.md
+++ b/README.md
@@ -174,9 +174,23 @@ determining that location is as follows:
 -   [gcloud sdk](https://cloud.google.com/sdk/install) >= 269.0.0
 -   [jq](https://stedolan.github.io/jq/) >= 1.6
 -   [Terraform](https://www.terraform.io/downloads.html) >= 0.12.6
--   [terraform-provider-google] plugin >= 2.1, < 4.0
--   [terraform-provider-google-beta] plugin >= 2.1, < 4.0
+-   [terraform-provider-google] plugin >= 3.1, < 4.0
+-   [terraform-provider-google-beta] plugin >= 3.1, < 4.0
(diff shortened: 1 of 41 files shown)


README.md
```diff
@@ -170,10 +174,24 @@ determining that location is as follows:
 -   [gcloud sdk](https://cloud.google.com/sdk/install) >= 269.0.0
 -   [jq](https://stedolan.github.io/jq/) >= 1.6
 -   [Terraform](https://www.terraform.io/downloads.html) >= 0.12.6
--   [terraform-provider-google] plugin >= 2.1, < 4.0
--   [terraform-provider-google-beta] plugin >= 2.1, < 4.0
+-   [terraform-provider-google] plugin >= 3.1, < 4.0
+-   [terraform-provider-google-beta] plugin >= 3.1, < 4.0
 -   [terraform-provider-gsuite] plugin 0.1.x if GSuite functionality is desired
 
+#### `terraform-provider-google` version 2.x
```
taylorludwig: I added a note in the readme about no longer supporting google provider 2.x and to use the previous version of this module if you need 2.x

But what's on master hasn't been pushed to terraform yet. We should push a release of what's in master before merging this so there is a 2.x version to use

modules/budget/main.tf
```diff
@@ -0,0 +1,63 @@
...
+
+resource "google_billing_budget" "budget" {
+  provider = google-beta
+  count    = var.create_budget ? 1 : 0
+
+  billing_account = var.billing_account
+  display_name    = local.display_name
+
+  budget_filter {
+    projects               = local.projects
```
taylorludwig: A bit strange, the provider docs say this only support a 0 or 1 projects array currently 
https://www.terraform.io/docs/providers/google/r/billing_budget.html#projects

However, passing in more than one works just fine. The test I created does just that. When setting more than one it also shows up in the billing budget UI on the console just fine too. So seems like the provider docs are just a bit out of date.

examples/budget_project/main.tf
```diff
@@ -0,0 +1,67 @@
...
+  folder_id         = var.folder_id
+  billing_account   = var.billing_account
+  budget_amount     = var.budget_amount
+}
+
+
+# An additional budget with more options
+resource "google_pubsub_topic" "budget" {
+  name    = "budget-topic-${random_string.suffix.result}"
+  project = var.pubsub_project_id
```
morgante: Just go ahead and put this in the project itself.

main.tf
```diff
@@ -57,4 +57,7 @@ module "project-factory" {
   disable_dependent_services  = var.disable_dependent_services
   python_interpreter_path     = var.python_interpreter_path
   pip_executable_path         = var.pip_executable_path
+  budget_amount               = var.budget_amount
```
morgante: I'd actually like to keep this *out* of the core project factory. Instead add a separate module invocation.

modules/budget/variables.tf
```diff
@@ -0,0 +1,66 @@
...
+}
+
+variable "projects" {
+  description = "The project ids to include in this budget. If empty budget will include all projects"
+  type        = list(string)
+}
+
+variable "amount" {
+  description = "The amount to use as the budget"
+  type        = string
```
morgante: Should this be a number?

modules/budget/variables.tf
```diff
@@ -0,0 +1,66 @@
...
+}
+
+variable "create_budget" {
+  description = "If the budget should be created"
+  type        = bool
+  default     = true
+}
+
+variable "display_name" {
+  description = "The display name of the budget"
```
morgante: Please add default behavior if unset to description.

modules/core_project_factory/main.tf
```diff
@@ -487,3 +487,17 @@ resource "google_project_iam_member" "gke_host_agent" {
     module.project_services,
   ]
 }
+
+/******************************************
```
morgante: Let's move this invocation into the respective top-level modules.

README.md
taylorludwig: Master was released as `6.2.1`. Updated readme to reflect

modules/budget/variables.tf
taylorludwig: Provider resource is a `string`. 
https://www.terraform.io/docs/providers/google/r/billing_budget.html#units

Maybe because it can be something other than USD

morgante: Weird, could we be more restrictive and mandate a number (which can be safely coerced to a string)?

modules/core_project_factory/main.tf
taylorludwig: Moved `budget` to be called from the root module directly

modules/budget/variables.tf
taylorludwig: Done

modules/budget/variables.tf
taylorludwig: Made it a `number` with a `tostring` call in the module

main.tf
taylorludwig: Moved it out and called budget submodule directly from root and other modules

examples/budget_project/main.tf
taylorludwig: Done

PR #361, opened 2020-01-28
Bump provider versions

wmuizelaar:
Also fixes the core_project_factory that still pointed to 2.x

### DIFF
--- a/examples/app_engine/main.tf
+++ b/examples/app_engine/main.tf
@@ -15,11 +15,11 @@
  */
 
 provider "google" {
-  version = "~> 3.3.0"
+  version = "~> 3.6.0"
 }
 
 provider "google-beta" {
-  version = "~> 3.3.0"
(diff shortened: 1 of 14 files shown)


morgante:
Thanks for the PR. What is the intent with this change? We try not to be overly strict with provider versions if we don't have to, so I'm not sure we need to restrict the provider versions further if we're not depending on a new feature.

wmuizelaar:
The main reason is the change in modules/fabric-project/versions.tf  - that one is needed to be able to use a 3.x.y provider version. Otherwise terraform will complain it could not find a valid version.

morgante:
Got it, thanks. In that case, could we reduce this change to update the examples + that file, while backing out the changes to other module versions?

wmuizelaar:
Sure! Hope to have applied your desired changes with this latest version.

PR #368, opened 2020-01-30
chore: Add a test fixture to validate dynamic Shared VPC

morgante:
This should prevent issues from occurring around the dynamically created/managed subnets or shared VPC host projects.

### DIFF
--- a/test/integration/dynamic_shared_vpc/controls/svpc.rb
+++ b/test/integration/dynamic_shared_vpc/controls/svpc.rb
@@ -1,0 +1,114 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+service_project_id          = attribute('service_project_id')
+service_project_number      = attribute('service_project_number')
+service_account_email       = attribute('service_account_email')
+shared_vpc                  = attribute('shared_vpc')
+shared_vpc_subnet_name_01   = attribute('shared_vpc_subnet_name_01')
+shared_vpc_subnet_region_01 = attribute('shared_vpc_subnet_region_01')
+shared_vpc_subnet_name_02   = attribute('shared_vpc_subnet_name_02')
+shared_vpc_subnet_region_02 = attribute('shared_vpc_subnet_region_02')
+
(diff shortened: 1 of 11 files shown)


morgante:
Issues/Todo:
- [ ] Update to 0.5 release of gcloud module
- [ ] Fix destroy order (service project needs to be destroyed first)

test/integration/dynamic_shared_vpc/controls/svpc.rb
```diff
@@ -0,0 +1,114 @@
...
+control 'project-factory-shared-vpc' do
+  title "Project Factory shared VPC"
+
+  describe command("gcloud compute shared-vpc get-host-project #{service_project_id} --format='get(name)'") do
+    its('exit_status') { should eq 0 }
+    its('stderr') { should eq '' }
+    its('stdout.strip') { should eq shared_vpc }
+  end
+
+  describe command("gcloud projects get-iam-policy #{shared_vpc} --format=json") do
```
rjerrems: Not a blockers, but It should be possible to do this directly in inspec:

https://www.inspec.io/docs/reference/resources/google_project_iam_bindings/

test/integration/dynamic_shared_vpc/controls/svpc.rb
```diff
@@ -0,0 +1,114 @@
...
+
+    it "includes the GKE service account in the roles/container.hostServiceAgentUser IAM binding" do
+      expect(bindings).to include(
+        members: including("serviceAccount:service-#{service_project_number}@container-engine-robot.iam.gserviceaccount.com"),
+        role: "roles/container.hostServiceAgentUser",
+      )
+    end
+  end
+
+  describe command("gcloud beta compute networks subnets get-iam-policy #{shared_vpc_subnet_name_01} --region #{shared_vpc_subnet_region_01} --project #{shared_vpc} --format=json") do
```
rjerrems: Similar to above https://www.inspec.io/docs/reference/resources/google_compute_subnetwork_iam_policy/

test/integration/dynamic_shared_vpc/controls/svpc.rb
```diff
@@ -0,0 +1,114 @@
...
+      it "includes the project service account in the roles/compute.networkUser IAM binding" do
+        expect(bindings).to include(
+          members: including("serviceAccount:#{service_account_email}"),
+          role: "roles/compute.networkUser",
+        )
+      end
+    end
+  end
+
+  describe command("gcloud beta compute networks subnets get-iam-policy #{shared_vpc_subnet_name_02} --region #{shared_vpc_subnet_region_02} --project #{shared_vpc} --format=json") do
```
rjerrems: Same as above https://www.inspec.io/docs/reference/resources/google_compute_subnetwork_iam_policy/

morgante:
> LGTM, only feedback is that we could use native inspec for some of the tests - we could track that as a separate issue as we have other tests that could be updated or is there a specific reason we are using gcloud commands directly?

At the time, gcloud commands were the only option. TBH, given available resourcing I don't think it's worth spending time/effort to switch over - they're functionally equivalent so no added value.

PR #369, opened 2020-01-31
[WIP] Add flag to skip gcloud download

### LINKED ISSUE #352: Disable automatic downloading of Cloud SDK
I have been using a terraform docker container with Cloud SDK and supporting libraries added to the image to perform my deployments. This has worked well in the past allowing me to ensure all required binaries and libraries are in place before deployment. This allowed me to peg the versions exactly as desired and deployments were generally pretty quick.

However it seems since 6.2.0, project factory tries to download all this for you during an apply. This has removed version control and has slowed down deployments significantly as Cloud SDK is getting downloaded every-time the project factory is been used where it really only needed to be installed once.

I can see the reason this was done, you want to ensure the right version of Cloud SDK is been used, but is it possible to disable this if desired and just rely on the user to install the Cloud SDK beforehand?

morgante:
Fixes #352


thesuperzapper:
@morgante any luck on this?

PR #387, opened 2020-03-02
Add dependency on service enablement.

### LINKED ISSUE #386: Race condition with adding bindings to GKE robot on new project
I've found a race condition when creating a new project using project-factory with the container.googleapis.com service, and adding role binding on the associated robot service account.

My case where it reproduces reliably is a little large, but the idea is the following:

```
module "project" {
  source                  = "terraform-google-modules/project-factory/google"
  version                 = "~> 7.0.0"
  random_project_id       = "false"
  name                    = format("mattcary-race-%s", random_id.suffix.hex)
  org_id                  = var.organization_id
  billing_account         = var.billing_account
  default_service_account = "keep"
  activate_apis           = ["compute.googleapis.com", "container.googleapis.com"]
}

resource "google_project_iam_member" "iam-binding" {
  project = module.project.project_id
  role    = "roles/container.developer"
  member  = "serviceAccount:service-${module.project.project_number}@container-engine-robot.iam.gserviceaccount.com"
}
```

If the above is done directly with google_project and google_project_service resources instead of the CFT module, then it also reproduces reliably.

Using the resources directly, I can put a depends_on in the iam-binding resource on the google-project_service resource to make it work reliably. 

Is there a way to do that using the project-factory module? For bonus points, what if I am making the binding on the robot account using a module? eg `terraform-google-modules/iam/google//modules/subnets_iam`.

Thanks

mattcary:
Adds a dependency on project services to the output project id and
number.  This prevents a race for using a robot account based on the
project id or number, as the robot accounts can be created only when
the service enabling is finished.

Fixes #386

### DIFF
--- a/modules/core_project_factory/outputs.tf
+++ b/modules/core_project_factory/outputs.tf
@@ -26,10 +26,12 @@ output "project_id" {
     ),
     0,
   )
+  depends_on = [ module.project_services ]
 }
 
 output "project_number" {
   value = google_project.main.number
+  depends_on = [ module.project_services ]
 }
 
 output "service_account_id" {
--- a/modules/core_project_factory/outputs.tf
+++ b/modules/core_project_factory/outputs.tf
@@ -26,12 +26,12 @@ output "project_id" {
     ),
     0,
   )
-  depends_on = [ module.project_services ]
+  depends_on = [module.project_services]
 }
 
 output "project_number" {
-  value = google_project.main.number
-  depends_on = [ module.project_services ]
+  value      = google_project.main.number
+  depends_on = [module.project_services]
 }
 
 output "service_account_id" {
--- a/test/fixtures/minimal/README.md
+++ b/test/fixtures/minimal/README.md
@@ -15,6 +15,7 @@
 | Name | Description |
 |------|-------------|
 | compute\_service\_account\_email |  |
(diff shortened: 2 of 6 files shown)


mattcary:
Fixes #386

mattcary:
I've run make docker_test_integration manually and everything was ok (except for the destroy step).

mattcary:
Thanks, done.

Should I try to add anything to testing? It looks like the `full` test fixture already enables the container services and mucks about with service accounts, so I could just add a project iam binding there.

But I'm not sure if you want to keep the tests more focused or whatever.

morgante:
Adding an additional IAM grant to the `full` fixture should be sufficient (and adding a test that verify it was granted).

mattcary:
Super. I'll do that and ping here when finished.

mattcary:
As discussed offline, I couldn't add tests to the full fixture as it isn't working. I added them to minimal instead (making it somewhat less minimal, but oh well...).

LMK if this isn't what you had in mind.

Thanks!

mattcary:
Done, thanks.

I can't see the cloud build output or errors, digging into the details I don't have a cloudbuild.builds.get permission. Is there something to be added to, or are there any other errors in the build?

Issue #391, opened 2020-03-04
gcloud_bin_abs_path changes on re-apply


dwilliams782:
Running a plan/apply on different machines produces the following diff:

```
# module.project_mobile.module.project-factory.module.gcloud_disable.null_resource.run_command[0] must be replaced
-/+ resource "null_resource" "run_command" {
      ~ id       = "9166635190702597732" -> (known after apply)
      ~ triggers = { # forces replacement
            "destroy_cmd_entrypoint"  = "gcloud"
          ~ "gcloud_bin_abs_path"     = "<local-path>.project-factory.gcloud_disable/terraform-google-modules-terraform-google-gcloud-09a6f10/cache/9272b146/google-cloud-sdk/bin" -> "<local-path>.project-factory.gcloud_disable/terraform-google-modules-terraform-google-gcloud-09a6f10/cache/9272b146/google-cloud-sdk/bin"
        }
    }

Plan: 1 to add, 0 to change, 1 to destroy.
```

Truncated the local paths, but it appears to be downloading the gcloud bin and storing the path in state, which isn't what we want to do. I noticed the skip_download variable but that isn't exposed to the project-factory module yet?

Apologies if it's a duplicate, I had a search and couldn't see anything.

morgante:
Unfortunately I don't think there's actually any way to get around this, as Terraform *requires* everything in the null_resource command to be a trigger and we need to know the gcloud path.

Fortunately, you can safely apply the local-exec multiple times (it's idempotent).

thesuperzapper:
@morgante isn't `self.triggers.gcloud_bin_abs_path` also stored as a `local.gcloud_bin_abs_path`?

Why cant we just remove `gcloud_bin_abs_path` from the triggers and use the the local variable in our local-exec?

morgante:
That's actually how it used to be: https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/pull/420/files

But then https://github.com/hashicorp/terraform/issues/23679 (in Terraform core) started warning/deprecating this behavior, so we can *only* access triggers.

dwilliams782:
> Fortunately, you can safely apply the local-exec multiple times (it's idempotent).

Whilst this is true, it muddies up our plans in our CI pipeline, so we have had to move away from using this module.

thesuperzapper:
Lets discuss this in #396

Luke-Vear:
> #396

[...]

PR #393, opened 2020-03-16
Add option for skipping the gcloud CLI download

MajorBreakfast:
Adds ability to skip the downloading of the gcloud CLI. Exposing this functionality is important to us because:
- the download functionality adds a lot of unnecessary overhead when running in our ci pipeline which runs in a container image with gcloud pre-installed
- the resources it adds to the terraform plan make the plan harder to read
- it seems to scale not so well with the amount of projects because the gcloud cli is installed once for every module instance instead of once overall

Similar to https://github.com/terraform-google-modules/terraform-google-project-factory/pull/369. However, that PR semms to contain other unreleated changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,8 @@ Extending the adopted spec, each change should have a link to its corresponding 
 
 ## [Unreleased]
 
+- Added `skip_gcloud_download` option which can be used to skip downloading the gcloud CLI if it is already installed
+
 ## [7.0.0] - 2020-01-15
 
 ### Fixed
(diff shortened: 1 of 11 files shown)


MajorBreakfast:
CI failed, but I don't have the permissions to see what went wront

bharathkkb:
@MajorBreakfast looks like it failed the linting step due to document generation. please run make docker_generate_docs and to test the lint step locally make docker_test_lint.

MajorBreakfast:
All green now. @bharathkkb Thanks for the instructions!

morgante:
Thanks for handling this!

thesuperzapper:
@MajorBreakfast, even with this change, I am still getting my changes polluted like #391.

If we are assuming `gcloud` will be in the PATH, cant we remove `gcloud_bin_abs_path` from the triggers of `null_resource.run_command`?

MajorBreakfast:
@thesuperzapper The usage of terraform-google-gcloud seems to create two problem. I've created an issue that summarizes them both here https://github.com/terraform-google-modules/terraform-google-project-factory/issues/396

adrian-gierakowski:
@MajorBreakfast is this just a placeholder for something that will be implemented in the future? terraform still seems to try to download gcloud even with `skip_gcloud_download = true`

PR #428, opened 2020-07-14
Feature/attach to vpc sc perimeter

kwraith05:
Adding ability to add project to an existing VPC service control perimeter (service perimeter attachment)

@bharathkkb  Please review and let me know of any changes I should make


bharathkkb:
@kwraith05 seems like there is some API enablement missing from the setup?

kwraith05:
@morgante  the logs for the build say the creation of the accesscontextmanager api project service was completed successfully, meaning the api should be enabled. However, the error says that if the API was enabled, it could take a few minutes for the action to propagate in the system and to wait a few minutes before trying again. 
@bharathkkb and I discussed and we're thinking we're getting this error because the project the master SA resides in doesn't have the API enabled. Do you know of any workaround?

bharathkkb:
Also @morgante for some additional context, the reason for moving `access_context_manager_policy` into the `setup` is because @kwraith05 ran into an issue with the `count value depends on resource attributes` error.

morgante:
> @bharathkkb and I discussed and we're thinking we're getting this error because the project the master SA resides in doesn't have the API enabled. Do you know of any workaround?

That's probably be the issue.

> Also @morgante for some additional context, the reason for moving access_context_manager_policy into the setup is because @kwraith05 ran into an issue with the count value depends on resource attributes error.

I don't quite follow. Why would we need to create it in `setup`? We should track down the actual reason for that error instead of moving it to the setup unnecessarily.

bharathkkb:
@morgante I believe the error originated from 
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/5a2be35a67014e8ff50682dcecf017e014301067/modules/core_project_factory/main.tf#L68

because value `var. vpc_service_control_perimeter_name` is not known until we actually create it. @kwraith05 could you confirm the issue and post logs if any?

morgante:
@bharathkkb If we're encountering that issue, I suggest exposing the parameter separately (`vpc_service_control_attach_enabled`) instead of trying to compute it. If we're seeing it, users will too.

bharathkkb:
Makes sense. @kwraith05 let me know if you need any help with the change recommended above.

bharathkkb:
@kwraith05 I think that error is stemming from https://github.com/terraform-providers/terraform-provider-google/issues/3582. As a fix can you bump it to latest wherever possible?

variables.tf
```diff
@@ -208,3 +208,9 @@ variable "skip_gcloud_download" {
   type        = bool
   default     = false
 }
+
```
bharathkkb, line 211: Wouldnt we need to expose `vpc_service_control_attach_enabled` as well and pass it in to core?

bharathkkb:
Also @kwraith05 I think the tests did not actually invoke the right code path as `vpc_service_control_attach_enabled` is set to `false` by default. In the fixture we need to set this to `true` after you do https://github.com/terraform-google-modules/terraform-google-project-factory/pull/428#discussion_r455983383

test/fixtures/vpc_sc_project/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+variable "parent_id" {
+  type        = string
+  description = "The parent of this AccessPolicy in the Cloud Resource Hierarchy. As of now, only organizations are accepted as parent."
+  default     = "11111111"
```
morgante, line 20: This is likely redundant with org_id? Let's remove if we don't need it.

test/fixtures/vpc_sc_project/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+}
+
+variable "policy_name" {
+  type        = string
+  description = "The policy's name"
+  default     = "policy1"
```
morgante, line 26: Remove this as it's not a sensible default.

test/fixtures/vpc_sc_project/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+  type        = string
+  description = "Billing account ID"
+}
+
+variable "random_string_for_testing" {
+  type        = string
+  description = "A random string of characters to be appended to resource names to ensure uniqueness"
+}
+
+variable "vpc_service_control_attach_enabled" {
```
morgante, line 49: This doesn't need to be a variable. The whole point of this fixture is to test VPCSC so we should hardcode to true.

test/fixtures/vpc_sc_project/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+  type        = bool
+  description = "Whether the project will be attached to a VPC Service Control Perimeter"
+  default     = false
+}
+
+variable "vpc_service_control_perimeter_name" {
```
morgante, line 55: Is this needed? We don't use it in the test.

test/fixtures/vpc_sc_project/variables.tf
```diff
@@ -0,0 +1,70 @@
...
+  description = "The name of an existing VPC Service Control Perimeter to add the created project to"
+  default     = null
+}
+
+variable "policy_id" {
+  type        = string
+  description = "The ID of the access context manager policy the perimeter lies in"
+}
+
+variable "service_account_email" {
```
morgante, line 66: Don't make this configurable, just leave sa_email off entirely.

test/integration/vpc_sc_project/controls/vpc_sc_project.rb
```diff
@@ -0,0 +1,91 @@
...
+  describe command("gcloud services list --project #{project_id}") do
+    its('exit_status') { should be 0 }
+    its('stderr') { should eq '' }
+
+    its('stdout') { should match(/compute\.googleapis\.com/) }
+    its('stdout') { should match(/accesscontextmanager\.googleapis\.com/) }
+    its('stdout') { should match(/storage-component.googleapis.com/) }
+  end
+
+  describe command("gcloud iam service-accounts list --project #{project_id} --format='json(email,disabled)'") do
```
morgante, line 50: No need to add these redundant tests.

test/integration/vpc_sc_project/controls/vpc_sc_project.rb
```diff
@@ -0,0 +1,91 @@
...
+# limitations under the License.
+
+project_id                         = attribute('project_id')
+project_number                     = attribute('project_number')
+service_account_email              = attribute('service_account_email')
+compute_service_account_email      = attribute('compute_service_account_email')
+group_email                        = attribute('group_email')
+group_name                         = attribute('group_name')
+
+control 'project-factory-vpc-sc-project' do
```
morgante, line 22: We should add a test which validates presence in the perimeter.

test/setup/iam.tf
```diff
@@ -57,6 +62,14 @@ resource "google_folder_iam_member" "int_test_folder" {
   member = "serviceAccount:${google_service_account.int_test.email}"
 }
 
+resource "google_organization_iam_member" "int_test_org" {
+  count = length(local.int_required_org_roles)
```
morgante, line 66: Use for_each instead of count.

bharathkkb:
@morgante I did it manually to unblock foundations, will send a PR soon.

test/integration/vpc_sc_project/controls/vpc_sc_project.rb
bharathkkb: Added

test/setup/iam.tf
bharathkkb: done

PR #434, opened 2020-07-30
Creating shared_vpc_access submodule

marko7460:
This submodule is used to control access of the API service accounts
on the host project. Service Accounts are given appropriate permissions
based on the list of active apis. The module is meant to be extended in
the future with more support for API service accounts.

Note: I have set `skip_gcloud_download  = "true"` in `examples/shared_vpc/main.tf` 
to speed up integration testing.

### DIFF
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -42,12 +42,13 @@ provider "random" {
   Host Project Creation
  *****************************************/
 module "host-project" {
-  source            = "../../"
-  random_project_id = true
-  name              = var.host_project_name
-  org_id            = var.organization_id
-  folder_id         = var.folder_id
-  billing_account   = var.billing_account
+  source               = "../../"
+  random_project_id    = true
+  name                 = var.host_project_name
+  org_id               = var.organization_id
+  folder_id            = var.folder_id
+  billing_account      = var.billing_account
+  skip_gcloud_download = true
 }
 
 /******************************************
@@ -119,9 +120,11 @@ module "service-project" {
   activate_apis = [
     "compute.googleapis.com",
     "container.googleapis.com",
+    "dataproc.googleapis.com",
   ]
 
   disable_services_on_destroy = "false"
+  skip_gcloud_download        = "true"
 }
 
 /******************************************
@@ -143,9 +146,11 @@ module "service-project-b" {
   activate_apis = [
     "compute.googleapis.com",
     "container.googleapis.com",
+    "dataproc.googleapis.com",
   ]
 
   disable_services_on_destroy = "false"
+  skip_gcloud_download        = "true"
 }
 
 /******************************************
--- a/modules/shared_vpc/main.tf
+++ b/modules/shared_vpc/main.tf
@@ -61,6 +61,17 @@ module "project-factory" {
 }
 
 /******************************************
+  Setting API service accounts for shared VPC
+ *****************************************/
+module "shared_vpc_access" {
+  source             = "../shared_vpc_access"
+  host_project_id    = var.shared_vpc
+  service_project_id = module.project-factory.project_id
+  active_apis        = module.project-factory.enabled_apis
+  shared_vpc_subnets = var.shared_vpc_subnets
+}
+
+/******************************************
   Billing budget to create if amount is set
  *****************************************/
 module "budget" {
--- a/modules/shared_vpc_access/README.md
+++ b/modules/shared_vpc_access/README.md
@@ -1,0 +1,41 @@
+# Shared VPC Access
+
+This module grants IAM permissions on host project and subnets to appropriate API service accounts based on activated
+APIs. For now only GKE and Dataproc APIs are supported.
+
+## Example Usage
+```hcl
+module "shared_vpc_access" {
+  source              = "terraform-google-modules/project-factory/google//modules/shared_vpc_access"
+  host_project_id     = var.shared_vpc
+  service_project_id  = var.service_project
+  active_apis         = [
+    "compute.googleapis.com",
+    "container.googleapis.com",
+    "dataproc.googleapis.com",
+  ]
+  shared_vpc_subnets  = [
+    "projects/pf-ci-shared2/regions/us-west1/subnetworks/shared-network-subnet-01",
+    "projects/pf-ci-shared2/regions/us-west1/subnetworks/shared-network-subnet-02",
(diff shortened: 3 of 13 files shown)


examples/shared_vpc/main.tf
```diff
@@ -42,12 +42,13 @@ provider "random" {
...
-  org_id            = var.organization_id
-  folder_id         = var.folder_id
-  billing_account   = var.billing_account
+  source               = "../../"
+  random_project_id    = true
+  name                 = var.host_project_name
+  org_id               = var.organization_id
+  folder_id            = var.folder_id
+  billing_account      = var.billing_account
+  skip_gcloud_download = "true"
```
bharathkkb, line 51: nit:
```suggestion
  skip_gcloud_download = true
```

modules/shared_vpc_access/outputs.tf
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+output "active_api_service_accounts" {
+  description = "List of active API service accounts in the service project."
+  value       = local.active_api_s_accounts
```
bharathkkb, line 19: I am wondering if this would cause issues with dependency ordering. My concern is that  `local.active_api_s_accounts` can be computed locally and doest not seem to dependent on any resources. Another thought is, we could output the project id and have a `depends_on` to enforce this.

marko7460: I do agree with you. Unfortunatelly `depends_on` block can not have variables in it. For example i can't do 
```hcl
output "something" {
  value = local.active_api_s_accounts
  depends_on = [var.project_id]
}```
The submodule doesn't need any outputs. I can just remove `outputs.tf`.

File: examples/shared_vpc/main.tf (hunk 1 above)
Comment (marko7460) [reply examples/shared_vpc/main.tf]: fixed

File: modules/shared_vpc_access/outputs.tf (hunk 2 above)
Comment (morgante) [reply modules/shared_vpc_access/outputs.tf:19]: That's not what a depends_on is meant to do - the point is to establish a connection with a resource.

Please add an output like this:
```
output "project_id" {
  value = var.project_id
  depends_on = [google_compute_subnetwork_iam_member.gke_shared_vpc_subnets, google_project_iam_member. gke_host_agent, google_project_iam_member. dataproc_shared_vpc_network_user]
}
```

Comment (marko7460) [reply modules/shared_vpc_access/outputs.tf:19]: added

File: modules/shared_vpc/main.tf [hunk 3]
```diff
@@ -60,6 +60,17 @@ module "project-factory" {
   skip_gcloud_download              = var.skip_gcloud_download
 }
 
+/******************************************
+  Setting API service accounts for shared VPC
+ *****************************************/
+module "shared_vpc_access" {
```
bharathkkb, line 66: I believe we should also modify the output for this module to
```hcl
output "project_id" {
  description = "If provided, the project uses the given project ID. Mutually exclusive with random_project_id being true."
  value       = module.shared_vpc_access.project_id
}
```
this way when this module is used, the project-id is returned only after `module.shared_vpc_access` is completed.

bharathkkb:
@marko7460 overall this lgtm. @morgante I believe this is a breaking change and we should have an upgrade guide as well?

modules/shared_vpc_access/README.md
```diff
@@ -0,0 +1,41 @@
+# Shared VPC Access
+
+This module sets IAM permissions on host project and host subnets to appripriate API service accounts based on activated
```
morgante, line 3:
```suggestion
This module grants IAM permissions on host project and subnets to appropriate API service accounts based on activated
```

modules/shared_vpc/main.tf
morgante: +1

marko7460:
I'll add a migration guide to `docs/upgrading_to_project_factory_9.0.md`

modules/shared_vpc/main.tf
marko7460: done

modules/shared_vpc_access/main.tf
```diff
@@ -0,0 +1,78 @@
...
+  project = var.host_project_id
+  role    = "roles/container.hostServiceAgentUser"
+  member  = format("serviceAccount:%s", local.gke_s_account)
+}
+
+/******************************************
+  compute.networkUser role granted to dataproc service account for dataproc on shared VPC subnets
+  See: https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/network#creating_a_cluster_that_uses_a_vpc_network_in_another_project
+ *****************************************/
+resource "google_project_iam_member" "dataproc_shared_vpc_network_user" {
```
jeffmccune, line 73: @marko7460 Shouldn't this be `google_compute_subnetwork_iam_member` to avoid granting access to all subnets in the host?

morgante: We'd need to support both paths, so adding another option like https://github.com/terraform-google-modules/terraform-google-project-factory/pull/434/files#diff-25530a290a058747ad4e0e2f230fc886R39

marko7460: The feature was coded based on https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/network#creating_a_cluster_that_uses_a_vpc_network_in_another_project. I did try giving permissions only on the subnet level and I was able to create a cluster only by specifying permissions on the subnet level. I can only confirm that those permissions are enough to create a cluster but i can't confirm if running and operating the cluster will not be an issue.

jeffmccune: @morgante What do you think about a future enhancement that breaks this one module into one module per API service, for the following reasons:

Consider two teams, red and blue, each wanting GKE and Dataproc.  There are 4 subnets, {red,blue}-{dataproc,gke}.  The purpose of the subnets are to prevent someone spinning up a cluster from impacting running clusters by exhausting IP space.

To support this use case a per-service module would take as input a list of subnets.  For each subnet it would manage a `google_compute_subnetwork_iam_member` resource with the service agent as a member and network user as a role.

Thoughts?

PR #436, opened 2020-08-03
Adding ability to tie subnets to API's

marko7460:
This commit enables users to specify if certain subnets are tied
to certains APIs. For example, now we can specify that a subnet is
a GKE subnet so that the roles/compute.networkUser will be granted
to GKE API service account on the subnet level.


morgante:
I'm not convinced we should do this. it's increasing the complexity substantially and I don't think is going to be useful for the majority of customers.

PR #438, opened 2020-08-12
bug: fix svpc regression

bharathkkb:
- fix svpc regression to grant `roles/compute.networkUser` at project level if subnets empty
- additional tests to cover all cases
    - if list of subnets empty, expect GKE/dataproc  bindings at project level (`service project b`)
    -  if list of subnets not empty, expect GKE/dataproc bindings at subnet level (`service project`)
    - Either case, if GKE,  `roles/container.hostServiceAgentUser` granted at project level
- use common resources for each service

### DIFF
--- a/modules/shared_vpc_access/main.tf
+++ b/modules/shared_vpc_access/main.tf
@@ -19,18 +19,14 @@ data "google_project" "service_project" {
 }
 
 locals {
-  gke_shared_vpc_enabled = contains(var.active_apis, "container.googleapis.com")
-  gke_s_account = local.gke_shared_vpc_enabled ? [format(
-    "service-%s@container-engine-robot.iam.gserviceaccount.com",
-    data.google_project.service_project.number,
-  )] : []
+  apis = {
+    "container.googleapis.com" : format("service-%s@container-engine-robot.iam.gserviceaccount.com", data.google_project.service_project.number),
+    "dataproc.googleapis.com" : format("service-%s@dataproc-accounts.iam.gserviceaccount.com", data.google_project.service_project.number),
+  }
+  gke_shared_vpc_enabled      = contains(var.active_apis, "container.googleapis.com")
   dataproc_shared_vpc_enabled = contains(var.active_apis, "dataproc.googleapis.com")
-  dataproc_s_account = local.dataproc_shared_vpc_enabled ? [format(
-    "service-%s@dataproc-accounts.iam.gserviceaccount.com",
-    data.google_project.service_project.number
-  )] : []
-  active_api_s_accounts = flatten([local.gke_s_account, local.dataproc_s_account])
-  subnetwork_api        = length(var.shared_vpc_subnets) != 0 ? setproduct(local.active_api_s_accounts, var.shared_vpc_subnets) : []
+  active_apis                 = setintersection(keys(local.apis), var.active_apis)
+  subnetwork_api              = length(var.shared_vpc_subnets) != 0 ? tolist(setproduct(local.active_apis, var.shared_vpc_subnets)) : []
 }
 
 /******************************************
@@ -54,7 +50,7 @@ resource "google_compute_subnetwork_iam_member" "gke_dataproc_shared_vpc_subnets
     index(split("/", local.subnetwork_api[count.index][1]), "regions") + 1,
   )
   project = var.host_project_id
-  member  = format("serviceAccount:%s", local.subnetwork_api[count.index][0])
+  member  = format("serviceAccount:%s", local.apis[local.subnetwork_api[count.index][0]])
 }
 
 /******************************************
@@ -62,10 +58,10 @@ resource "google_compute_subnetwork_iam_member" "gke_dataproc_shared_vpc_subnets
  if "dataproc.googleapis.com" compute.networkUser role granted to dataproc service account for dataproc on shared VPC Project if no subnets defined
  *****************************************/
 resource "google_project_iam_member" "gke_dataproc_shared_vpc_network_user" {
-  count   = length(var.shared_vpc_subnets) == 0 && (local.gke_shared_vpc_enabled || local.dataproc_shared_vpc_enabled) ? length(local.active_api_s_accounts) : 0
-  project = var.host_project_id
-  role    = "roles/compute.networkUser"
-  member  = format("serviceAccount:%s", local.active_api_s_accounts[count.index])
+  for_each = length(var.shared_vpc_subnets) == 0 && (local.gke_shared_vpc_enabled || local.dataproc_shared_vpc_enabled) ? local.active_apis : []
+  project  = var.host_project_id
+  role     = "roles/compute.networkUser"
+  member   = format("serviceAccount:%s", local.apis[each.value])
 }
 
 /******************************************
@@ -76,6 +72,5 @@ resource "google_project_iam_member" "gke_host_agent" {
   count   = local.gke_shared_vpc_enabled ? 1 : 0
   project = var.host_project_id
   role    = "roles/container.hostServiceAgentUser"
-  member  = format("serviceAccount:%s", local.gke_s_account[0])
+  member  = format("serviceAccount:%s", local.apis["container.googleapis.com"])
 }
-
--- a/modules/shared_vpc_access/main.tf
+++ b/modules/shared_vpc_access/main.tf
@@ -20,16 +20,17 @@ data "google_project" "service_project" {
 
 locals {
   gke_shared_vpc_enabled = contains(var.active_apis, "container.googleapis.com")
-  gke_s_account = local.gke_shared_vpc_enabled ? format(
+  gke_s_account = local.gke_shared_vpc_enabled ? [format(
     "service-%s@container-engine-robot.iam.gserviceaccount.com",
     data.google_project.service_project.number,
-  ) : ""
+  )] : []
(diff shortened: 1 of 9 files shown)


modules/shared_vpc_access/main.tf
```diff
@@ -20,16 +20,17 @@ data "google_project" "service_project" {
...
+  )] : []
   dataproc_shared_vpc_enabled = contains(var.active_apis, "dataproc.googleapis.com")
-  dataproc_s_account = local.dataproc_shared_vpc_enabled ? format(
+  dataproc_s_account = local.dataproc_shared_vpc_enabled ? [format(
     "service-%s@dataproc-accounts.iam.gserviceaccount.com",
     data.google_project.service_project.number
-  ) : ""
-  active_api_s_accounts = compact([local.gke_s_account, local.dataproc_s_account])
+  )] : []
+  active_api_s_accounts = flatten([local.gke_s_account, local.dataproc_s_account])
```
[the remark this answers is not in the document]
morgante: I think looping on the actual accounts is actually dangerous. I think something like this would be better:

```
locals {
  apis = {
    "container.googleapis.com": format("service-%s@container-engine-robot.iam.gserviceaccount.com", data.google_project.service_project.number
 }
  active_apis = setintersection(keys(local.apis), var.active_apis
  subnetwork_api = setproduct(local.active_apis, var.shared_vpc_subnets)
}
```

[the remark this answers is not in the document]
bharathkkb: I ended up having to do `tolist(setproduct(local.active_apis, var.shared_vpc_subnets))` as it was otherwise a set and throwing error `This value does not have any indices`.  In this case I dont think element ordering matters as it becomes a list of elements of type [api,subnet].

modules/shared_vpc_access/main.tf
```diff
@@ -19,40 +19,49 @@ data "google_project" "service_project" {
...
-  compute.networkUser role granted to GKE service account for GKE on shared VPC subnets
+  if "container.googleapis.com" compute.networkUser role granted to GKE service account for GKE on shared VPC subnets
+  if "dataproc.googleapis.com" compute.networkUser role granted to dataproc service account for dataproc on shared VPC subnets
   See: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc
  *****************************************/
-resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
+resource "google_compute_subnetwork_iam_member" "gke_dataproc_shared_vpc_subnets" {
   provider = google-beta
-  count    = local.gke_shared_vpc_enabled && length(var.shared_vpc_subnets) != 0 ? length(var.shared_vpc_subnets) : 0
+  count    = length(var.shared_vpc_subnets) != 0 && (local.gke_shared_vpc_enabled || local.dataproc_shared_vpc_enabled) ? length(local.subnetwork_api) : 0
```
morgante, line 39: Can we change this to `for_each = local.subnetwork_api`?

modules/shared_vpc_access/main.tf
```diff
@@ -19,40 +19,49 @@ data "google_project" "service_project" {
...
-  gke_shared_vpc_enabled = contains(var.active_apis, "container.googleapis.com")
-  gke_s_account = local.gke_shared_vpc_enabled ? format(
-    "service-%s@container-engine-robot.iam.gserviceaccount.com",
-    data.google_project.service_project.number,
-  ) : ""
+  apis = {
+    "container.googleapis.com" : format("service-%s@container-engine-robot.iam.gserviceaccount.com", data.google_project.service_project.number),
+    "dataproc.googleapis.com" : format("service-%s@dataproc-accounts.iam.gserviceaccount.com", data.google_project.service_project.number),
+  }
+  gke_shared_vpc_enabled      = contains(var.active_apis, "container.googleapis.com")
```
morgante, line 26: We shouldn't need these individual _enabled locals.

modules/shared_vpc_access/main.tf
```diff
@@ -63,16 +72,5 @@ resource "google_project_iam_member" "gke_host_agent" {
...
-/******************************************
-  compute.networkUser role granted to dataproc service account for dataproc on shared VPC subnets
-  See: https://cloud.google.com/dataproc/docs/concepts/configuring-clusters/network#creating_a_cluster_that_uses_a_vpc_network_in_another_project
- *****************************************/
-resource "google_project_iam_member" "dataproc_shared_vpc_network_user" {
-  count   = local.dataproc_shared_vpc_enabled ? 1 : 0
-  project = var.host_project_id
-  role    = "roles/compute.networkUser"
-  member  = format("serviceAccount:%s", local.dataproc_s_account)
+  member  = format("serviceAccount:%s", local.apis["container.googleapis.com"])
```
morgante, line 75: We should drop the explicit per-API resources?

modules/shared_vpc_access/main.tf
bharathkkb: yeah I tried this yesteday, but ran into the `cannot determine number of resources error`. I will give it another whirl with new structure.

bharathkkb: So I think the problem with for_each is that at plan time we only know the length and not the values. Similar to this issue here https://github.com/hashicorp/terraform/issues/22735#issuecomment-580491563
Let me know if you have other suggestions.

I tried something like
```hcl
locals{
subnetwork_api = [for pair in setproduct(var.active_apis, var.shared_vpc_subnets): {
    api = pair[0]
    subnet = pair[1]
    }
  ] : []
}
 resource "google_compute_subnetwork_iam_member" "gke_dataproc_shared_vpc_subnets" {
...
for_each    = {for pair in local.subnetwork_api: "${pair.api}.${pair.subnet}" => pair}
}
```

morgante: But we should know the values of the APIs though, no?

bharathkkb: Yeah I think its the subnets_self_links since they are created dynamically
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/9eb64e2217bef8477ae07e1a834c4bfb3f64273f/examples/shared_vpc/main.tf#L118

modules/shared_vpc_access/main.tf
bharathkkb: We used to create this in `8.x` 
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/af1db49139c30199cd80b39eb6369d704f3ef151/modules/core_project_factory/main.tf#L469
and this is needed for GKE svpc?

modules/shared_vpc_access/main.tf
bharathkkb: This check will be needed for the `hostServiceAgentUser` binding but if that can be removed, we can get rid of this as well.

modules/shared_vpc_access/main.tf
morgante: Right but let's do this in a loop for each API (same approach as we used for subnets).

modules/shared_vpc_access/main.tf
morgante: Hmm ok I guess we can stick to count then.

modules/shared_vpc_access/main.tf
morgante: Right but why do we need a conditional for dataproc? We should *only* use `gke_shared_vpc_enabled` for the hostServiceAgentUser part.

bharathkkb: Oops didnt see dataproc_enabled was still there, will clean up :)

modules/shared_vpc_access/main.tf
```diff
@@ -19,40 +19,48 @@ data "google_project" "service_project" {
...
 }
 
 /******************************************
-  compute.networkUser role granted to GKE service account for GKE on shared VPC subnets
+  if "container.googleapis.com" compute.networkUser role granted to GKE service account for GKE on shared VPC subnets
+  if "dataproc.googleapis.com" compute.networkUser role granted to dataproc service account for dataproc on shared VPC subnets
   See: https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc
  *****************************************/
-resource "google_compute_subnetwork_iam_member" "gke_shared_vpc_subnets" {
+resource "google_compute_subnetwork_iam_member" "gke_dataproc_shared_vpc_subnets" {
```
morgante, line 36:
```suggestion
resource "google_compute_subnetwork_iam_member" "service_shared_vpc_subnet_users" {
```

modules/shared_vpc_access/main.tf
```diff
@@ -19,40 +19,48 @@ data "google_project" "service_project" {
...
   project = var.host_project_id
-  member  = format("serviceAccount:%s", local.gke_s_account)
+  member  = format("serviceAccount:%s", local.apis[local.subnetwork_api[count.index][0]])
+}
+
+/******************************************
+ if "container.googleapis.com" compute.networkUser role granted to GKE service account for GKE on shared VPC Project if no subnets defined
+ if "dataproc.googleapis.com" compute.networkUser role granted to dataproc service account for dataproc on shared VPC Project if no subnets defined
+ *****************************************/
+resource "google_project_iam_member" "gke_dataproc_shared_vpc_network_user" {
```
morgante, line 59:
```suggestion
resource "google_project_iam_member" "service_shared_vpc_user" {
```

Issue #445, opened 2020-08-18
shared_vpc features is broken (or removed) from root module


askoriy:
After upgrade to version 9.0 I could not recreate shared_vpc resources that I had in 8.1 version.
[Examples](https://github.com/terraform-google-modules/terraform-google-project-factory#usage), Documentation, [Upgrading guide](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/v9.0.0/docs/upgrading_to_project_factory_v9.0.md#gke-api-already-enabled) - nothing helped.
Finally, I found the solution is to use `shared_vpc` submodule, guessed by https://github.com/terraform-google-modules/terraform-google-project-factory/issues/444

Please add back shared_vpc features to root module, as it was previously or update the documentation to make it obvious for people who upgrade to 9.0 to use `//modules/shared_vpc` submodule when upgrading to 9.0 if shared_vpc networks are used.

morgante:
@askoriy Is there a reason you can't switch to the shared_vpc submodule? I think simply documenting this is going to be a better approach.

des-esseintes:
Faced the same issue.
What's the proper way of bringing back shared_vpcs?

morgante:
@des-esseintes You should use the `shared_vpc` submodule.

des-esseintes:
thanks @morgante!
hmmmm this assumes moving resources in terraform state... damn, i hate it :) was just thinking maybe there's an easier path. nevermind :)

morgante:
@des-esseintes If you switch from the root module to the shared VPC submodule there shouldn't be any major issues as the state location is mostly the same. If you do so then share your Terraform plan I'm happy to tell you if it's a breaking change.

des-esseintes:
it worked well, thanks!

askoriy:
Submodule `//modules/shared_vpc` works fine if it's known that the project is _service project_ and both `shared_vpc` and `shared_vpc_subnets` variables provided with non-empty values. But it tails if one of them is empty, therefore submodule can't be used as universal project factory.

I could not reproduce errors from [this](https://github.com/terraform-google-modules/terraform-google-project-factory/pull/446#issuecomment-705607111) comment, but changes in https://github.com/terraform-google-modules/terraform-google-project-factory/pull/446 made root module works for me again for both _host_ and _service_ projects, so I created fork until fix test errors and merge that PR

bharathkkb:
[...]

PR #446, opened 2020-08-18
fix: Add shared_vpc features to root module back

askoriy:
This PR fixes https://github.com/terraform-google-modules/terraform-google-project-factory/issues/445


morgante (dismissed):

askoriy:
@morgante could you point to build error details to finally fix it and merge this PR?

main.tf
```diff
@@ -64,6 +64,17 @@ module "project-factory" {
   vpc_service_control_perimeter_name = var.vpc_service_control_perimeter_name
 }
 
+/******************************************
+  Setting API service accounts for shared VPC
+ *****************************************/
+module "shared_vpc_access" {
+  source             = "./modules/shared_vpc_access"
+  host_project_id    = var.shared_vpc
```
bharathkkb, line 72: Would we not need additional checks in `shared_vpc_access` to catch the  case if `var.shared_vpc == ""`, which is the default value? I am also concerned if those checks (if implemented with `coun`t) would bring a scenario where TF may complain about being unable to determine number of resources until run time.

askoriy:
@morgante or @bharathkkb could you point me to error details?

bharathkkb:
@askoriy 
PFA
```
       
         on ../../../modules/shared_vpc_access/main.tf line 64, in resource "google_project_iam_member" "service_shared_vpc_user":
         64:   for_each = (length(var.shared_vpc_subnets) == 0) && (var.host_project_id != "") ? local.active_apis : []
       
       The "for_each" value depends on resource attributes that cannot be determined
       resources that the for_each depends on.
```

bharathkkb:
@morgante as discussed here https://github.com/terraform-google-modules/terraform-google-project-factory/issues/445#issuecomment-675824866 are we planning on adding this back to the root module or updating docs to reflect usage of shared_vpc submodule?

morgante:
I'm inclined to update docs so it's only available with the shared_vpc submodule.

morgante:
> @morgante as discussed here [#445 (comment)](https://github.com/terraform-google-modules/terraform-google-project-factory/issues/445#issuecomment-675824866) are we planning on adding this back to the root module or updating docs to reflect usage of shared_vpc submodule?

So long as we can get it to work properly, we should add it back in this release. If we want to remove it totally, that should be clearly communicated in the next major release.

modules/shared_vpc_access/main.tf
```diff
@@ -61,7 +61,7 @@ resource "google_compute_subnetwork_iam_member" "service_shared_vpc_subnet_users
    if "dataflow.googleapis.com" compute.networkUser role granted to dataflow  service account for Dataflow on shared VPC Project if no subnets defined
  *****************************************/
 resource "google_project_iam_member" "service_shared_vpc_user" {
-  for_each = length(var.shared_vpc_subnets) == 0 ? local.active_apis : []
+  for_each = (length(var.shared_vpc_subnets) == 0) && (var.host_project_id != "") ? local.active_apis : []
```
bharathkkb, line 64: I believe what is happening here is in our svpc example https://github.com/terraform-google-modules/terraform-google-project-factory/blob/v8.1.0/examples/shared_vpc/main.tf  we test dynamic host + service project creation. However when you added `var.host_project_id != ""`, `var.host_project_id` is determined dynamically and hence TF complains ` "for_each" value depends on resource attributes that cannot be determined until apply`.

I believe one solution could be modifying the `shared_vpc_access` to take in a boolean `shared_vpc_enabled` variable. Then in the `shared_vpc` module change the `shared_vpc_access` module instantiation to 
```tf
module "shared_vpc_access" {
  ...
  shared_vpc_enabled = true
}
```

In the root module change the `shared_vpc_access` module instantiation to 
```tf
module "shared_vpc_access" {
  ...
  shared_vpc_enabled = var.host_project_id != "" ? true : false
}
```
 This will enable users who want to dynamically provision both host and service projects to use a combination of the root module (for host project) and  `shared_vpc` module (for service projects) as shown [here](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/examples/shared_vpc/main.tf) 

/cc @morgante would this preserve historical behavior for both the root and `shared_vpc` modules?

morgante: Hi @bharathkkb I think that approach makes sense.

des-esseintes:
hey @askoriy do you still think this stuff has to be merged or maybe you found some workaround to make life easier with the existing functionality?

morgante (approved):

PR #449, opened 2020-08-27
Upgrade gcloud module to 2.0.0

Dev25:
https://github.com/terraform-google-modules/terraform-google-gcloud/releases/tag/v2.0.0

>The module has been changed to not download gcloud by default. You can override this behavior by setting skip_download = false or setting the GCLOUD_TF_DOWNLOAD environment variable to "always".

~Also setting `skip_gcloud_download` to true by default~

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -151,7 +151,7 @@ data "null_data_source" "default_service_account" {
  *****************************************/
 module "gcloud_delete" {
   source  = "terraform-google-modules/gcloud/google"
-  version = "~> 1.0.0"
+  version = "~> 2.0.0"
 
   enabled                           = var.default_service_account == "delete"
   use_tf_google_credentials_env_var = var.use_tf_google_credentials_env_var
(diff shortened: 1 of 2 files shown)


morgante:
@bharathkkb @Dev25 What do you think of dropping the skip variables entirely? Now that we have the env-based approach I think it's much more robust.

Dev25:
I would be in favour of that @morgante
Do you know if the docker image already sets that var so we can rely on that for CI?

morgante:
> Do you know if the docker image already sets that var so we can rely on that for CI?

With the 2.0 version, the default is now to skip so we can rely on that.

bharathkkb:
@morgante I agree as well. TF cloud also seems to let you set env vars, so users should be able to use that. However in the CI imho we should still test the codepath of enabling via `GCLOUD_TF_DOWNLOAD` for one of the int tests.

morgante:
@bharathkkb We test the override here - https://github.com/terraform-google-modules/terraform-google-gcloud/tree/master/test/fixtures/override_example

PR #450, opened 2020-08-31
feat: Add service identity

### LINKED ISSUE #448: Support service_identity
Googled managed API service accounts are being moved to a just-in-time activation rather than on API enablement. This breaks many Terraform based workflows which create projects, enable APIs and grants additional IAM permissions to them.

A new `google_project_service_identity` resource is being created which can force create the SA (see https://github.com/hashicorp/terraform-provider-google/issues/7020).

I believe the best place to call this resource is project factory. Would you accept a contribution to add field `activate_service_identitities` alongisde `activate_apis` to this module? This will in turn create the identities for the listed APIs.

We could go a step further and automatically append `activate_service_identities` to `activate_apis`. I can't think of a case where you would want to create an API SA but not enable the API itself.

umairidris:
Fix #448 

This won't work till google-beta provider is released with the service identity resource next Monday.

### DIFF
--- a/modules/project_services/outputs.tf
+++ b/modules/project_services/outputs.tf
@@ -23,3 +23,8 @@ output "enabled_apis" {
   description = "Enabled APIs in the project"
   value       = [for api in google_project_service.project_services : api.service]
 }
+
+output "enabled_api_identities" {
+  description = "Enabled API identities in the project"
+  value       = {for i in google_project_service_identity.project_service_identities : i.service => i.email}
+}
--- a/modules/project_services/outputs.tf
+++ b/modules/project_services/outputs.tf
@@ -26,5 +26,5 @@ output "enabled_apis" {
 
 output "enabled_api_identities" {
   description = "Enabled API identities in the project"
-  value       = {for i in google_project_service_identity.project_service_identities : i.service => i.email}
+  value       = { for i in google_project_service_identity.project_service_identities : i.service => i.email }
 }
--- a/README.md
+++ b/README.md
@@ -109,6 +109,7 @@ determining that location is as follows:
 
 | Name | Description | Type | Default | Required |
 |------|-------------|:----:|:-----:|:-----:|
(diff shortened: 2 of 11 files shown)


modules/project_services/outputs.tf
```diff
@@ -23,3 +23,8 @@ output "enabled_apis" {
   description = "Enabled APIs in the project"
   value       = [for api in google_project_service.project_services : api.service]
 }
+
+output "enabled_api_identities" {
+  description = "Enabled API identities in the project"
+  value       = { for i in google_project_service_identity.project_service_identities : i.service => i.email }
```
bharathkkb, line 29: @umairidris can we modify these hardcoded identities to use this now?
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/099cdcce28642b045337b4ca0a0c54a9949d9285/modules/shared_vpc_access/main.tf#L23

umairidris: The challenge with using this resource is you will need to manually grant any roles that were previously automatically granted on activation (e.g. for healthcare SA the healthcare.serviceAgent role is automatically granted).

So if using this resource it should be fine but you will also need to ensure you grant any permissions that you were previously assuming were automatically granted, else it will not have the right permissions to do its job.

morgante: I think we might want to consider adding functionality for setting IAM roles on SAs to Project Factory, with sensible defaults which match the old defaults.

umairidris: Could you elaboroate?

Currently the input from user looks something like this:

```
activate_api_identities = [{
    api = "healthcare.googleapis.com"
    roles = [
      "roles/healthcare.serviceAgent",
      "roles/bigquery.jobUser",
    ]
  }]
```

Do you want us to add additional helpers to set roles/healthcare.serviceAgent by default? If so, should we manually maintain a mapping of API to default role? I am afraid that might be inconsistent if we miss certain APIs, though https://cloud.google.com/iam/docs/understanding-roles#service-management-roles might help. I'll start an internal thread with the team on potential automation.

morgante: Got it, I think we can merge as-is and evolve based on the thread.

morgante (approved):

PR #454, opened 2020-09-10
[DRAFT] chore: parallel builds for int tests

bharathkkb:
Recently CI for PF seems to be failing at the destroy step. For instance #450 
I have been unable to repro locally but my theory is that it is due to multiple gcloud destroys happening in the same build step. This aims to build them in separate build steps in parallel. 

edit: that alone did not solve the problem completely, ended up skipping gcloud for all but one example

Changelist
- parallel builds
- added global `GCLOUD_TF_DOWNLOAD=always`
- set `GCLOUD_TF_DOWNLOAD=never` for minimal example


morgante:
@bharathkkb Please ping me once tests are passing / you're ready for review.

bharathkkb:
@morgante looks like this has stabilized and  we have some savings in build time too. ~45m -> ~15m.
Updated the description with all changes.
/cc @umairidris

morgante (approved):

Issue #455, opened 2020-09-11
Add monitoring_notification_channels to modules/budget


davidholsgrove:
Provider issue https://github.com/hashicorp/terraform-provider-google/issues/6868 added support for `monitoring_notification_channels` in `google_billing_budget` 

Update the `modules/budget` to support this option, inline with the documented example: https://www.terraform.io/docs/providers/google/r/billing_budget.html#example-usage-billing-budget-notify

PR #456, opened 2020-09-14
Add budget notification channel

davidholsgrove:
Allow configuring billing budget with monitoring notification channels.

Fixes https://github.com/terraform-google-modules/terraform-google-project-factory/issues/455


davidholsgrove:
Hi @adrienthebo / @morgante - any chance of a review on this PR?
The budget alerts through Monitoring Notification channel email appear to be more useful out-of-the-box than the pub/sub method, due to the budget publishing messages multiple times a day, not just at budget threshold.

morgante:
Thanks for the contribution! @davidholsgrove Do you think we should accept the email and create the channel inline within the module?

davidholsgrove:
> 
> 
> Thanks for the contribution! @davidholsgrove Do you think we should accept the email and create the channel inline within the module?

Good question - I followed the approach in the module currently, where we don't create the pub/sub resources, and just take the topic name as a var.

I think creating the notification channel in-module might be more difficult, as in practice if people follow https://cloud.google.com/monitoring/workspaces#monitored-projects the monitoring workspace project will be different from the budget project

morgante (approved):

morgante:
@davidholsgrove Good point, I think this is fine as is. We could adjust this later if need be.

PR #458, opened 2020-09-17
feat: Add Dataflow to Shared VPC API service accounts

zefdelgadillo:
Add Dataflow API service account to Shared VPC host project if API is enabled on new projects. see https://cloud.google.com/dataflow/docs/concepts/security-and-permissions#cloud_dataflow_service_account


bharathkkb (approved):
LGTM
Thanks for the PR!

Issue #422, opened 2020-05-26
Enable shared_vpc_host to be set using project factory 


bharathkkb:
Currently inorder to create a project as a SVPC host project, we have to either
- Create project using PF and then use `google_compute_shared_vpc_host_project` to mark it as SVPC host
- [Use Network module ](https://github.com/terraform-google-modules/terraform-google-network)

We should consider wrapping `google_compute_shared_vpc_host_project` within PF similar to [how we attach as a service project.](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/9da81586f218c13c69f13a4034f606203ac8187c/modules/core_project_factory/main.tf#L136)

morgante:
I think this is a reasonable addition. We could expose an `enable_shared_vpc_host` variable.

thiagonache:
we can close this issue, can't we?

PR #459, opened 2020-09-23
Address Issue #422

thiagonache:
This PR provides an option to create the project and set it as a shared vpc host project.
It should address #422.

@bharathkkb @morgante lmk if that's what you guys had in mind.

### DIFF
--- a/examples/svpc_host_project/README.md
+++ b/examples/svpc_host_project/README.md
@@ -1,0 +1,34 @@
+# Shared VPC Host Project
+
+This example illustrates how to create a [Shared VPC](https://cloud.google.com/vpc/docs/shared-vpc) host project.
+
+It includes creating the host project and using the [network module](https://github.com/terraform-google-modules/terraform-google-network) to create network.
+
+It then attaches two new service projects to the host project.
+
+<!-- BEGINNING OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
+## Inputs
+
+| Name | Description | Type | Default | Required |
+|------|-------------|:----:|:-----:|:-----:|
+| billing\_account | The ID of the billing account to associate this project with | string | n/a | yes |
+| folder\_id | The folder to create projects in | string | n/a | yes |
+| host\_project\_name | Name for Shared VPC host project | string | `"shared-vpc-host"` | no |
+| network\_name | Name for Shared VPC network | string | `"shared-network"` | no |
+| organization\_id | The organization id for the associated services | string | n/a | yes |
+| service\_project\_name | Name for Shared VPC service project | string | `"shared-vpc-service"` | no |
+
+## Outputs
+
+| Name | Description |
+|------|-------------|
+| host\_project | The full host project info |
+| host\_project\_id | The ID of the created project |
+| network\_name | The name of the VPC being created |
+| network\_self\_link | The URI of the VPC being created |
+| service\_project | The service project info |
+| service\_project\_b | The second service project |
+| subnets | The shared VPC subets |
+| vpc | The network info |
+
+<!-- END OF PRE-COMMIT-TERRAFORM DOCS HOOK -->
--- a/examples/svpc_host_project/README.md
+++ b/examples/svpc_host_project/README.md
@@ -26,7 +26,6 @@ It then attaches two new service projects to the host project.
 | host\_project\_id | The ID of the created project |
 | network\_name | The name of the VPC being created |
 | network\_self\_link | The URI of the VPC being created |
-| service\_project | The service project info |
 | service\_project\_b | The second service project |
 | subnets | The shared VPC subets |
 | vpc | The network info |
--- a/examples/svpc_host_project/README.md
+++ b/examples/svpc_host_project/README.md
@@ -26,6 +26,7 @@ It then attaches two new service projects to the host project.
 | host\_project\_id | The ID of the created project |
 | network\_name | The name of the VPC being created |
 | network\_self\_link | The URI of the VPC being created |
+| service\_project | The first service project |
 | service\_project\_b | The second service project |
 | subnets | The shared VPC subets |
 | vpc | The network info |
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -137,6 +137,12 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
   ]
 }
 
+resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
+  count      = var.enable_shared_vpc_host ? 1 : 0
+  project    = google_project.main.project_id
+  depends_on = [google_project.main]
+}
+
 /******************************************
   Default compute service account retrieval
  *****************************************/
--- a/modules/svpc_host_project/main.tf
+++ b/modules/svpc_host_project/main.tf
@@ -1,0 +1,87 @@
+/**
+ * Copyright 2018 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*****************************************
+  Organization info retrieval
+ *****************************************/
+module "gsuite_group" {
+  source = "../gsuite_group"
+
+  domain = var.domain
+  name   = var.group_name
+  org_id = var.org_id
+}
+
+module "project-factory" {
+  source = "../core_project_factory"
+
+  group_email                       = module.gsuite_group.email
+  group_role                        = var.group_role
+  lien                              = var.lien
+  manage_group                      = var.group_name != "" ? "true" : "false"
+  random_project_id                 = var.random_project_id
+  org_id                            = var.org_id
+  name                              = var.host_project_name
+  project_id                        = var.project_id
+  shared_vpc                        = var.shared_vpc
+  enable_shared_vpc_host            = true
+  shared_vpc_enabled                = false
+  billing_account                   = var.billing_account
+  folder_id                         = var.folder_id
+  sa_role                           = var.sa_role
+  activate_apis                     = var.activate_apis
+  usage_bucket_name                 = var.usage_bucket_name
+  usage_bucket_prefix               = var.usage_bucket_prefix
+  bucket_versioning                 = var.bucket_versioning
+  credentials_path                  = var.credentials_path
+  shared_vpc_subnets                = var.shared_vpc_subnets
+  labels                            = var.labels
+  bucket_project                    = var.bucket_project
+  bucket_name                       = var.bucket_name
+  bucket_location                   = var.bucket_location
+  auto_create_network               = var.auto_create_network
+  disable_services_on_destroy       = var.disable_services_on_destroy
+  default_service_account           = var.default_service_account
+  disable_dependent_services        = var.disable_dependent_services
+  python_interpreter_path           = var.python_interpreter_path
+  use_tf_google_credentials_env_var = var.use_tf_google_credentials_env_var
+  skip_gcloud_download              = var.skip_gcloud_download
+}
+
+# /******************************************
+#   Setting API service accounts for shared VPC
+#  *****************************************/
+# module "shared_vpc_access" {
+#   source             = "../shared_vpc_access"
+#   host_project_id    = var.shared_vpc
+#   service_project_id = module.project-factory.project_id
+#   active_apis        = module.project-factory.enabled_apis
+#   shared_vpc_subnets = var.shared_vpc_subnets
+# }
+
+/******************************************
+  Billing budget to create if amount is set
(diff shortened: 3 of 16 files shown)


modules/core_project_factory/main.tf
```diff
@@ -137,6 +137,12 @@ resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
   ]
 }
 
+resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
+  count      = var.enable_shared_vpc_host ? 1 : 0
+  project    = google_project.main.project_id
+  depends_on = [google_project.main]
```
bharathkkb, line 143: We should probably be depending on `module.project_services` to make sure that compute api is enabled first?

modules/svpc_host_project/main.tf
```diff
@@ -0,0 +1,87 @@
+/**
```
bharathkkb, line 1: I am not sure about the purpose of this module?  IIUC we just need to make the addition in core and plumb it to the root module.

examples/svpc_host_project/README.md
```diff
@@ -0,0 +1,34 @@
+# Shared VPC Host Project
```
bharathkkb, line 1: We can probably condense this into the [existing SVPC example](https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/examples/shared_vpc). I am thinking removing [this flag ](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/9bc317e763f767d5666f6876fdae91b3e9a6b200/examples/shared_vpc/main.tf#L65) and ideally this addition should take care of activating the host project. WDYT?

bharathkkb (changes requested):
Thanks for the PR! @thiagonache

modules/core_project_factory/main.tf
thiagonache: You're right!

modules/svpc_host_project/main.tf
thiagonache: So, my idea is to give two examples, but it does not mean I need two modules :) 
Going to fix.

examples/svpc_host_project/README.md
thiagonache: My initial idea is to have both examples just to avoid introducing a breaking change. Like, the ppl using the current example won't need to bother. Please, lmk your thoughts.

thiagonache:
@bharathkkb @morgante I'm refactoring this PR to work as you guys want but I'm not able to figure out when the example should point to a module or to the root dir. For example, the shared_vpc inside of examples points to ../../ which is the root dir even there is a module in ../../modules/shared_vpc.
I'm asking that because [the current logic](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/62f5b091306134902e8c961ba7e3322e8e7429d1/main.tf#L40) in the root repo must be changed in order to expose just one variable since we have three different cases (host project, service project, and project without vpc). If I can point to the module I won't need to change variables and logic in the main root.

Please, let me know.

examples/svpc_host_project/README.md
bharathkkb: @thiagonache It would not be a breaking change as it is not at the module level. We do not expect end users to directly use the example other than perhaps copy pasting.

thiagonache: Got it.

thiagonache:
> @bharathkkb @morgante I'm refactoring this PR to work as you guys want but I'm not able to figure out when the example should point to a module or to the root dir. For example, the shared_vpc inside of examples points to ../../ which is the root dir even there is a module in ../../modules/shared_vpc.
> I'm asking that because [the current logic](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/62f5b091306134902e8c961ba7e3322e8e7429d1/main.tf#L40) in the root repo must be changed in order to expose just one variable since we have three different cases (host project, service project, and project without vpc). If I can point to the module I won't need to change variables and logic in the main root.
> 
> Please, let me know.

@bharathkkb what about it? The logic to attach the vpc is if the shared_vpc variable is not empty, this way it's always going to try to attach the host vpc as a service project from itself which won't work.

bharathkkb:
> the shared_vpc inside of examples points to ../../ which is the root dir even there is a module in ../../modules/shared_vpc

this is because we wrap `core_project_factory` with `shared_vpc_enabled  = true` along with `shared_vpc_access` to provide additional features if used as a service project. The example does [refer](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/62f5b091306134902e8c961ba7e3322e8e7429d1/examples/shared_vpc/main.tf#L107) to `../modules/shared_vpc` for its service projects.

I think the we can use an approach similar to https://github.com/terraform-google-modules/terraform-google-project-factory/pull/428/files and we expose `enable_shared_vpc_host` at the root module and is passed to core. Your [current implementation](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2b70001afe398124035d4dfbf8058da898012185/modules/core_project_factory/main.tf) in core lgtm. This will not be a breaking change as default behavior is false, which is the current behavior.

/cc @morgante if any other suggestions.

bharathkkb:
>  The logic to attach the vpc is if the shared_vpc variable is not empty

Yeah we may need to tweak that [logic](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/62f5b091306134902e8c961ba7e3322e8e7429d1/modules/core_project_factory/main.tf#L129) to take `enable_shared_vpc_host` into account.
Perhaps something like
```
resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
  count = var.shared_vpc_enabled  && !var. enable_shared_vpc_host ? 1 : 0
}
resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
  count = var.shared_vpc_enabled  && var. enable_shared_vpc_host ? 1 : 0
}
```

This way by default we have original behavior, i.e attach as service project but if `enable_shared_vpc_host` is true then become the  host project?

thiagonache:
> Your [current implementation](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2b70001afe398124035d4dfbf8058da898012185/modules/core_project_factory/main.tf) in core lgtm. This will not be a breaking change as default behavior is false, which is the current behavior.

The problem with it is [here](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2b70001afe398124035d4dfbf8058da898012185/modules/core_project_factory/main.tf#L130). Unless I add `&& ! enable_shared_vpc_host` in the condition it will try to create the attachment of the host project as a service project. And my concern about adding that and not expression is that confusing with the way we set [shared_vpc_enable](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2b70001afe398124035d4dfbf8058da898012185/main.tf#L40)
Let me know what do you think or if I'm missing something.

thiagonache:
> > The logic to attach the vpc is if the shared_vpc variable is not empty
> 
> Yeah we may need to tweak that [logic](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/62f5b091306134902e8c961ba7e3322e8e7429d1/modules/core_project_factory/main.tf#L129) to take `enable_shared_vpc_host` into account.
> Perhaps something like
> 
> ```
> resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
>   count = var.shared_vpc_enabled  && !var. enable_shared_vpc_host ? 1 : 0
> }
> resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
>   count = var.shared_vpc_enabled  && var. enable_shared_vpc_host ? 1 : 0
> }
> ```
> 
> This way by default we have original behavior, i.e attach as service project but if `enable_shared_vpc_host` is true then become the host project?

yeah... if we are okay with that it should do the job

bharathkkb:
>  And my concern about adding that and not expression is that confusing with the way we set shared_vpc_enable

Gotcha, thanks for catching that. I do see the problem with the logic by which we determine `var.shared_vpc_enabled`

What if we change root to do
```hcl
module "project-factory" {
  source = "./modules/core_project_factory"
  shared_vpc  = var.shared_vpc
  enable_shared_vpc_service  = var.shared_vpc != "" && !var.enable_shared_vpc_host 
  enable_shared_vpc_host = var.enable_shared_vpc_host
```
this way `enable_shared_vpc_service` is true iff `var.shared_vpc` has a value and `var.enable_shared_vpc_host` is false (default value for enable_shared_vpc_host is false, so it preserves current interface)
and in core
```hcl
resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
  count = var.enable_shared_vpc_service ? 1 : 0
}
resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
  count = var. enable_shared_vpc_host ? 1 : 0
}
```
An edge case would be if `var. shared_vpc` is "foo" and `var.enable_shared_vpc_host` is true then `var.enable_shared_vpc_service` is false and we make it a host _This is a user error and  var.shared_vpc is not used however the user is explicitly setting `enable_shared_vpc_host` is true so makes sense we make it a host_

WDYT? Would this satisfy our use cases?

thiagonache:
> > And my concern about adding that and not expression is that confusing with the way we set shared_vpc_enable
> 
> Gotcha, thanks for catching that. I do see the problem with the logic by which we determine `var.shared_vpc_enabled`
> 
> What if we change root to do
> 
> ```terraform
> module "project-factory" {
>   source = "./modules/core_project_factory"
>   shared_vpc  = var.shared_vpc
>   enable_shared_vpc_service  = var.shared_vpc != "" && !var.enable_shared_vpc_host 
>   enable_shared_vpc_host = var.enable_shared_vpc_host
> ```
> 
> this way `enable_shared_vpc_service` is true iff `var.shared_vpc` has a value and `var.enable_shared_vpc_host` is false (default value for enable_shared_vpc_host is false, so it preserves current interface)
> and in core
> 
> ```terraform
> resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
>   count = var.enable_shared_vpc_service ? 1 : 0
> }
> resource "google_compute_shared_vpc_host_project" "shared_vpc_host" {
>   count = var. enable_shared_vpc_host ? 1 : 0
> }
> ```
> 
> An edge case would be if `var. shared_vpc` is "foo" and `var.enable_shared_vpc_host` is true then `var.enable_shared_vpc_service` is false and we make it a host _This is a user error and var.shared_vpc is not used however the user is explicitly setting `enable_shared_vpc_host` is true so makes sense we make it a host_
> 
> WDYT? Would this satisfy our use cases?

That's exactly what I had in mind, two variables one for shared vpc service and another one for shared vpc host. Cool.

I'll refactor like that and open a new PR from a clean branch. I'll keep this open until I have the new PR and can do a reference.

Thank you!

morgante:
IMO we don't need to overcomplicate this. A few notes:

1. In `core_project_factory` we should change the `shared_vpc_enabled` variable to be `enable_shared_vpc_service_project`.
2. In `modules/shared_vpc`, change it to force `enabled_shared_vpc_service_project`.
3. In `core_project_factory` add a new `enable_shared_vpc_host_project` variable and plumb that through to the root module as well.
4. There is no need for a separate `shared_vpc_host` submodule: the root module can satisfy that use case. The only reason we have a separate service project submodule is for cases with dynamically generated network names.

Anything I missed?

thiagonache:
> IMO we don't need to overcomplicate this. A few notes:
> 
> 1. In `core_project_factory` we should change the `shared_vpc_enabled` variable to be `enable_shared_vpc_service_project`.
> 2. In `modules/shared_vpc`, change it to force `enabled_shared_vpc_service_project`.
> 3. In `core_project_factory` add a new `enable_shared_vpc_host_project` variable and plumb that through to the root module as well.
> 4. There is no need for a separate `shared_vpc_host` submodule: the root module can satisfy that use case. The only reason we have a separate service project submodule is for cases with dynamically generated network names.
> 
> Anything I missed?

No. I over-complicated in the beginning because I did completely understand how things were composed. Sorry. It's my first time contributing to this project.
You basically resumed what @bharathkkb and I agreed. So, thank you for confirming we are in the right path now.

thiagonache:
> > IMO we don't need to overcomplicate this. A few notes:
> > 
> > 1. In `core_project_factory` we should change the `shared_vpc_enabled` variable to be `enable_shared_vpc_service_project`.
> > 2. In `modules/shared_vpc`, change it to force `enabled_shared_vpc_service_project`.
> > 3. In `core_project_factory` add a new `enable_shared_vpc_host_project` variable and plumb that through to the root module as well.
> > 4. There is no need for a separate `shared_vpc_host` submodule: the root module can satisfy that use case. The only reason we have a separate service project submodule is for cases with dynamically generated network names.
> > 
> > Anything I missed?
> 
> No. I over-complicated in the beginning because I did not completely understand how things were composed. Sorry. It's my first time contributing to this project.
> You basically resumed what @bharathkkb and I agreed. So, thank you for confirming we are on the right path now.

BTW when I said two variables I meant renaming one and adding another one.

morgante:
@thiagonache Great! Thank you for contributing. 🎉

thiagonache:
> @thiagonache Great! Thank you for contributing. 🎉

@morgante @bharathkkb One last thing, should we rename shared_vpc to host_project_id? The name shared_vpc is meaningless after many interactions on the module. Even when you read the description seems disconnected IMO
```
variable "shared_vpc" {
  description = "The ID of the host project which hosts the shared VPC"
  type        = string
  default     = ""
}
```

thiagonache:
> @thiagonache I'd rather not force a breaking change like that, we can keep it for our next major release (#463).

perfect!

thiagonache:
closing in favor of PR https://github.com/terraform-google-modules/terraform-google-project-factory/pull/464

PR #464, opened 2020-10-01
Create project as host project (shared_vpc_host)

thiagonache:
Hi @morgante @bharathkkb ,

After a lot of discussions, I'm opening a new PR with a clean branch. I hope all is correct now.

### DIFF
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -42,13 +42,15 @@ provider "random" {
   Host Project Creation
  *****************************************/
 module "host-project" {
-  source               = "../../"
-  random_project_id    = true
-  name                 = var.host_project_name
-  org_id               = var.organization_id
-  folder_id            = var.folder_id
-  billing_account      = var.billing_account
-  skip_gcloud_download = true
+  source                         = "../../"
+  random_project_id              = true
+  name                           = var.host_project_name
+  org_id                         = var.organization_id
+  folder_id                      = var.folder_id
+  billing_account                = var.billing_account
+  skip_gcloud_download           = true
+  shared_vpc                     = var.host_project_name
+  enable_shared_vpc_host_project = true
 }
 
 /******************************************
@@ -58,11 +60,9 @@ module "vpc" {
   source  = "terraform-google-modules/network/google"
   version = "~> 2.1.0"
 
-  project_id   = module.host-project.project_id
-  network_name = var.network_name
-
+  project_id                             = module.host-project.project_id
+  network_name                           = var.network_name
   delete_default_internet_gateway_routes = true
-  shared_vpc_host                        = true
 
   subnets = [
     {
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -109,12 +109,12 @@ module "service-project" {
   name              = var.service_project_name
   random_project_id = "false"
 
-  org_id             = var.organization_id
-  folder_id          = var.folder_id
-  billing_account    = var.billing_account
-  shared_vpc_enabled = true
+  org_id                            = var.organization_id
+  folder_id                         = var.folder_id
+  billing_account                   = var.billing_account
+  enable_shared_vpc_service_project = true
 
-  shared_vpc         = module.vpc.project_id
+  shared_vpc         = module.host-project.project_id
   shared_vpc_subnets = module.vpc.subnets_self_links
 
   activate_apis = [
@@ -137,12 +137,12 @@ module "service-project-b" {
   name              = "b-${var.service_project_name}"
   random_project_id = "false"
 
-  org_id             = var.organization_id
-  folder_id          = var.folder_id
-  billing_account    = var.billing_account
-  shared_vpc_enabled = true
+  org_id                            = var.organization_id
+  folder_id                         = var.folder_id
+  billing_account                   = var.billing_account
+  enable_shared_vpc_service_project = true
 
-  shared_vpc = module.vpc.project_id
+  shared_vpc = module.host-project.project_id
 
   activate_apis = [
     "compute.googleapis.com",
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -121,6 +121,7 @@ module "service-project" {
     "compute.googleapis.com",
     "container.googleapis.com",
     "dataproc.googleapis.com",
+    "dataflow.googleapis.com",
   ]
 
   disable_services_on_destroy = "false"
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -49,7 +49,6 @@ module "host-project" {
   folder_id                      = var.folder_id
   billing_account                = var.billing_account
   skip_gcloud_download           = true
-  shared_vpc                     = var.host_project_name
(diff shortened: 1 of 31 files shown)


thiagonache:
Not sure why it's failing running make on my local machine it does pass.

morgante:
The error is here:
```
terraform_validate ./test/fixtures/full 

Error: Unsupported argument

  on .terraform/modules/vpc/main.tf line 58, in resource "google_compute_subnetwork" "subnetwork":
  58:   enable_flow_logs         = lookup(var.subnets[count.index], "subnet_flow_logs", "false")

An argument named "enable_flow_logs" is not expected here.
```

thiagonache:
I don’t think it is related to my changes... do you?

thiagonache:
> I don’t think it is related to my changes... do you?
> […](#)
> On 1 Oct 2020, at 16:56, Morgante Pell ***@***.***> wrote: ﻿ The error is here: terraform_validate ./test/fixtures/full Error: Unsupported argument on .terraform/modules/vpc/main.tf line 58, in resource "google_compute_subnetwork" "subnetwork": 58: enable_flow_logs = lookup(var.subnets[count.index], "subnet_flow_logs", "false") An argument named "enable_flow_logs" is not expected here. — You are receiving this because you authored the thread. Reply to this email directly, view it on GitHub, or unsubscribe.

What I've found is that enable_flow_logs was deprecated, we should use log_config instead. Trying to match the same version that is working for me.

thiagonache:
@morgante for some odd reason the google provider version pulled by my terraform execution end up with a different version than in the CI environment. I've deleted the terraform cache and confirmed that it pulled out the version 3.40 again, so I've created the required version config for the provider in fixtures/versions.tf and it worked.
The last thing I've just done is to revert the commit to pin the module version that was my initial suspicion. I'm confident that it will work this time.

thiagonache:
yay!

examples/shared_vpc/main.tf
```diff
@@ -42,13 +42,15 @@ provider "random" {
...
-  billing_account      = var.billing_account
-  skip_gcloud_download = true
+  source                         = "../../"
+  random_project_id              = true
+  name                           = var.host_project_name
+  org_id                         = var.organization_id
+  folder_id                      = var.folder_id
+  billing_account                = var.billing_account
+  skip_gcloud_download           = true
+  shared_vpc                     = var.host_project_name
```
morgante, line 52: We shouldn't need this on the host project.

modules/core_project_factory/variables.tf
```diff
@@ -182,11 +182,17 @@ variable "disable_dependent_services" {
...
 
-variable "shared_vpc_enabled" {
-  description = "If shared VPC should be used"
+variable "enable_shared_vpc_service_project" {
+  description = "If this project should be attached to a shared VPC. If true, you must set shared_vpc variable"
   type        = bool
 }
 
+variable "enable_shared_vpc_host_project" {
+  description = "If this project is a shared VPC host project. If true, you must set shared_vpc variable. Default is false for compatibility"
```
morgante, line 191: Actually if it's the host project we should *not* set `shared_vpc` variable.

modules/gsuite_enabled/variables.tf
```diff
@@ -171,8 +171,8 @@ variable "disable_dependent_services" {
   type        = string
 }
 
-variable "shared_vpc_enabled" {
-  description = "If shared VPC should be used"
+variable "enable_shared_vpc_service_project" {
```
morgante, line 174: Please leave the name unchanged on outer modules for backwards-compatibility.

modules/shared_vpc/variables.tf
```diff
@@ -165,8 +165,8 @@ variable "disable_dependent_services" {
   type        = bool
 }
 
-variable "shared_vpc_enabled" {
-  description = "If shared VPC should be used"
+variable "enable_shared_vpc_service_project" {
```
morgante, line 168: Please leave the name unchanged on outer modules for backwards-compatibility.

variables.tf
```diff
@@ -48,6 +48,12 @@ variable "shared_vpc" {
   default     = ""
 }
 
+variable "enable_shared_vpc_host_project" {
+  description = "If this project is a shared VPC host project. If true, you must set shared_vpc variable. Default is false for compatibility"
```
morgante, line 52:
```suggestion
  description = "If this project is a shared VPC host project. If true, you must set *not* shared_vpc variable. Default is false."
```

morgante (changes requested):
When creating the host project you should *not* need a `shared_vpc` input variable with the name.

thiagonache:
> When creating the host project you should _not_ need a `shared_vpc` input variable with the name.

that's the problem with the logic and mentioned. We agreed on that...
the only way to determine the variable is if shared_vpc is set

variables.tf
thiagonache: @morgante actually, no... shared_vpc is the name of the host vpc, this is a boolean.

modules/shared_vpc/variables.tf
thiagonache: @morgante you suggested renaming this variable. You meant we should have two different names in different places for the same variable? I'm confused now.

modules/gsuite_enabled/variables.tf
thiagonache: again, you asked to rename this variable. should I do it only in the root module? it's going to be confusing.

modules/core_project_factory/variables.tf
thiagonache: we must set... remember that shared_vpc == host project name.. that why we opened the issue to rename this variable

examples/shared_vpc/main.tf
thiagonache: again, shared vpc is the host project name... it's better to repeat here than add an if in the code

thiagonache:
@morgante that's the current code
```
shared_vpc                         = var.shared_vpc
shared_vpc_enabled                 = var.shared_vpc != ""
```
is your suggestion is just to change to
```
shared_vpc                                          = var.shared_vpc
enable_shared_vpc_service_project                 = var.shared_vpc != ""
enable_shared_vpc_host_project = var. enable_shared_vpc_host_project
```
This is a lot confusing IMO
Please, confirm.

morgante:
> that's the problem with the logic and mentioned. We agreed on that...
> the only way to determine the variable is if shared_vpc is set

No it's not. There is no reason to set `shared_vpc` on the host project—it already *is* the host project. Just ignore it completely. It's very confusing to pass a name to the `shared_vpc` argument on the host project.

thiagonache:
> > that's the problem with the logic and mentioned. We agreed on that...
> > the only way to determine the variable is if shared_vpc is set
> 
> No it's not. There is no reason to set `shared_vpc` on the host project—it already _is_ the host project. Just ignore it completely. It's very confusing to pass a name to the `shared_vpc` argument on the host project.

yeah... I've removed this comment. Please, see my last comment

morgante:
> @morgante okay. you're right, but I think we should rename shared_vpc_enabled on the next major release as we will do with shared_vpc, do you agree?

Yep.

thiagonache:
Please, see related PR.

PR #475, opened 2020-10-09
chore: update desc of activate_api_identities

umairidris:


umairidris:
@bharathkkb something a bit odd... when i run `sudo make generate_docs` it doesnt seem to update the readmes. Any idea why?

bharathkkb:
@umairidris could be https://github.com/terraform-docs/terraform-docs/issues/105, looks like our image has 0.6.0, so we may need to bump it.

umairidris:
Good catch. Sent https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/814

I think we will need to wait for 0.13 migration? The makefile is pinned to tag 0.12

morgante:
Note that it could also be related to possible issues in our wrapper script: https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/blob/master/infra/build/developer-tools/build/scripts/task_wrapper_scripts/terraform_docs.sh

bharathkkb:
@umairidris after debugging through the GKE module CI failures I think we should also add the `active_api_identities` to [this example ](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/examples/gke_shared_vpc) to showcase how to restore `roles/container.serviceAgent`.  We can add a line or two to the [FAQ](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/docs/FAQ.md#how-does-the-project-factory-work-with-a-shared-vpc) part explaining the behavior and link to this example. This can also be something we do within our `shared_vpc_access` submodule for GKE and a select few to be transparent to the user in the future or pending a datasource automate it completely.

modules/core_project_factory/variables.tf
```diff
@@ -90,7 +90,12 @@ variable "activate_apis" {
 }
 
 variable "activate_api_identities" {
-  description = "The list of service identities (Google Managed service account for the API) to force-create for the project (e.g. in order to grant additional roles). APIs in this list will automatically be appended to `activate_apis`. Not including the API in this list will follow the default behaviour for identity creation (which is usually when the first resource using the API is created)."
+  description = <<<EOF
```
bharathkkb, line 93: @umairidris I actually just noticed shouldn't this be 
```suggestion
  description = <<EOF
```

umairidris: D'oh! Good catch. It still isn't passing the docs generation though :(

morgante:
@bharathkkb @umairidris Is this PR ready for merge?

umairidris:
I have regenerated docs with the latest image. PTAL

morgante (approved):

bharathkkb (changes requested):
@umairidris we need to update here too for Cloud Build
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2d7fcb474c4492b2fd785e1631352dfcc7ead597/build/lint.cloudbuild.yaml#L24

umairidris:
> @umairidris we need to update here too for Cloud Build
> https://github.com/terraform-google-modules/terraform-google-project-factory/blob/2d7fcb474c4492b2fd785e1631352dfcc7ead597/build/lint.cloudbuild.yaml#L24

Done.

bharathkkb (approved):

Issue #407, opened 2020-04-27
Remove preconditions script


morgante:
In order to remove the [preconditions script](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/scripts/preconditions/preconditions.py) from the module (#373), we need to confirm it's impossible to start module creation and be left in a catastrophic state.

Therefore we need to make sure each step has an associated in-provider check and we document/explain in the troubleshooting guide.

Checks:
- [ ] Seed project - check that APIs are active
  * [x] [cloudresourcemanager.googleapis.com](#missing-api-cloudresourcemanagergoogleapiscom)
  * [x] [cloudbilling.googleapis.com](#missing-api-cloudbillinggoogleapiscom) - fixed via https://github.com/terraform-providers/terraform-provider-google/pull/5671
  * [x] [iam.googleapis.com](#missing-api-iamgoogleapiscom)
  * [x] [appengine.googleapis.com](#missing-appenginegoogleapiscom) (when managing an App Engine instance on the Target Project)
  * [x] [admin.googleapis.com](#missing-admingoogleapiscom) (when managing G Suite
  groups and group membership)
  * [x] `serviceusage.googleapis.com` - needed to remove default network - https://github.com/terraform-providers/terraform-provider-google/issues/6222
- [x] Confirm permissions on billing account
- [x] Confirm permissions on Shared VPC
- [x] Confirm permissions on organization
- [x] Confirm permissions on folder

thiagonache:
@morgante I know it's silly, but what if we force serviceusage to be enabled until we get it handled by the provider?
```
variable "activate_apis" {
  description = "The list of apis to activate within the project"
  type        = list(string)
  default     = ["compute.googleapis.com"]

  validation {
    condition     = can(lookup(var.activate_apis, "serviceusage.googleapis.com"))
    error_message = "Service Usage API is required due to network auto creation."
  }
}
```

Just a thought....

morgante:
@thiagonache It actually doesn't need to be enabled on the project being created, it needs to be on the credentials project.

thiagonache:
Doh!
Do you think they would accept the new resource to remove the default network? If yes, I can see if I can do it. I think I can.

thiagonache:
Hi @morgante! We got a small change merged that handles the serviceusage API. I know it's not your preferred way due to other possible errors, but I think if we tackle the most recurrent ones is enough to get rid of the preconditions script.

[...]

PR #478, opened 2020-10-14
Removes preconditions.py script into the module

thiagonache:
This PR removes the null_resource for preconditions.py script as per the issue https://github.com/terraform-google-modules/terraform-google-project-factory/issues/407


test/scripts/preconditions/test_preconditions.py
```diff
@@ -1,153 +0,0 @@
-#!/usr/bin/env python3
```
morgante, line 1: We shouldn't delete the tests.

docs/running_preconditions_script.md
```diff
@@ -0,0 +1,79 @@
...
+    ```
+    Check #1 (Required APIs on service account project) => It is missing to enable admin, cloudresourcemanager and cloudbilling services APIs in the <b>my-seed-project</b>.
+
+    Check #2 (Service account permissions on billing accoun) => The permission required to associate projects with billing accounts is okay.
+
+    Check #3 (Service account permissions on parent folder) => The permission to create new projects into the folder specified is granted.
+
+    Check #4 (Service account permissions on organization) => No permission required since we are creating the project under the folder instead of the organisation. If no folder is specified it would be step three and require projects.create permission.
+
+    You can add one last check by setting --shared-vpc parameter.
```
morgante, line 78:
```suggestion
    You can add one last check by setting the `--shared-vpc` parameter.
```

morgante (changes requested):
Thanks for working on this!

test/scripts/preconditions/test_preconditions.py
thiagonache: right... I should understand it first. Sorry.

thiagonache:
> Thanks for working on this!

Sorry for the delay. I've been busy. I wanted to do this past week.

morgante (approved):

iestarks:
Sorry if this is not the right area to raise this question, but I have been trying to run this and even using the setup script, I am still experiencing this issue and not sure what IAM member and role to adjust to get past this error.

Any ideas?

Error: failed pre-requisites: missing permission on "billingAccounts/012C34-27AF51-73A832": billing.resourceAssociations.create

morgante:
@iestarks You need to make sure your service account (or user identity) has Billing User on the Billing Account - https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/docs/TROUBLESHOOTING.md#missing-rolesbillinguser-role

PR #483, opened 2020-10-23
Support optionally grant GKE SA network admin on VPC host project

xingao267:
Fixes https://github.com/terraform-google-modules/terraform-google-kubernetes-engine/issues/723

### DIFF
--- a/modules/shared_vpc_access/variables.tf
+++ b/modules/shared_vpc_access/variables.tf
@@ -36,7 +36,7 @@ variable "active_apis" {
   default     = []
 }
 
-variable "gke_shared_vpc_security_admin_role_enabled" {
+variable "grant_services_security_admin_role" {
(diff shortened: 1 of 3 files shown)


modules/shared_vpc_access/variables.tf
```diff
@@ -35,3 +35,9 @@ variable "active_apis" {
   type        = list(string)
   default     = []
 }
+
+variable "gke_shared_vpc_security_admin_role_enabled" {
```
morgante, line 39: We could probably make this slightly more generic as future services might also need it.

```suggestion
variable "grant_services_security_admin_role" {
```

xingao267: Done.

morgante (approved):

Issue #484, opened 2020-10-28
failed to install provider (gsuite)


haf:
```
$ tf init
Initializing modules...

Initializing the backend...

Initializing provider plugins...
- Using previously-installed hashicorp/external v2.0.0
- Using previously-installed hashicorp/google v3.45.0
- Using previously-installed hashicorp/google-beta v3.45.0
- Using previously-installed hashicorp/null v2.1.2
- Using previously-installed hashicorp/random v2.3.1
- Finding hashicorp/gsuite versions matching "~> 0.1"...

Error: Failed to install provider

Error while installing hashicorp/gsuite: provider registry
registry.terraform.io does not have a provider named
registry.terraform.io/hashicorp/gsuite
```

I've searched the issue tracker and found only older questions about this, all related to versions of terraform and versions of this module, since upgraded. How do I initialise this module to a working state with gsuite support?

I'm following https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/gsuite_enabled/README.md (which seems to have the version wrong, too)

bharathkkb:
@haf Since this is a 3rd party provider, I think they only recently started publishing this with Terraform 0.13. Can you try
```hcl
terraform {
  required_providers {
    gsuite = {
      source = "DeviaVir/gsuite"
      version = "0.1.54"
    }
  }
}

provider "gsuite" {
...
}
```

haf:
That seems to work. Updated docs. Closing.

PR #485, opened 2020-10-29
Document required_providers when used with GSuite

haf:
Ref #484


morgante (changes requested):
We should add this into the module itself.

bharathkkb:
@morgante if we include the `source  = namespace/provider` in the module, [we will need to version constraint TF >0.12.26](https://www.terraform.io/docs/configuration/provider-requirements.html#v0-12-compatible-provider-requirements). 

Additionally, as  0.12.26+ ignores the `source  = namespace/provider`, I think this will fail anyway with a 3rd party provider.

PR #491, opened 2020-11-03
Replaces gcloud commands by terraform resource

thiagonache:


bharathkkb (commented):
Thanks for the PR and the work in the provider @thiagonache 

I assume we can also get rid of  the associated vars `skip_gcloud_download` `use_tf_google_credentials_env_var` etc and make this a breaking release. We would also need an upgrade guide to go along with it.

thiagonache:
> Thanks for the PR and the work in the provider @thiagonache
> 
> I assume we can also get rid of the associated vars `skip_gcloud_download` `use_tf_google_credentials_env_var` etc and make this a breaking release. We would also need an upgrade guide to go along with it.

Great! I'll work on it soon.

morgante:
+1, this will need to be a breaking change w/ an upgrade guide which explains how to delete the `gcloud` local-exec resources entirely.

thiagonache:
I think we'll need to have another migrate.py script because just terraform state rm won't work. I may be missing something, but the state will change relatively as far as I can see. Can someone just give a second thought to make sure I'm not missing anything? @morgante @bharathkkb

thiagonache:
Please, **hold on**. I was missing one item (data)

bharathkkb:
@thiagonache if you have an initial version of the upgrade guide, I am happy to test it.

thiagonache:
@bharathkkb you read my mind... just pushed it.

thiagonache:
did you get a chance to test it out? just wondering if you have any doubts. I've assumed that the new resource will be released on the next release (3.47).

morgante:
@thiagonache I have some concerns about this deleting all SAs which have the role on existing projects. IMO we should actually filter which SAs are deleted (as we do in the script).

thiagonache:
> @thiagonache I have some concerns about this deleting all SAs which have the role on existing projects. IMO we should actually filter which SAs are deleted (as we do in the script).

@morgante I'm not sure if I'm following. Are you talking about the upgrade guide?

bharathkkb:
@thiagonache I will test it out soon, but looking at the instructions do we have to remove it from state manually? I don't see any gcloud destroy_cmd so it will be no-op. Without the manual `terraform state rm ` commands did it give an error when you tried to upgrade?

thiagonache:
Yes. The transform step generates crash trying to access the state

thiagonache:
> > @thiagonache I have some concerns about this deleting all SAs which have the role on existing projects. IMO we should actually filter which SAs are deleted (as we do in the script).
> 
> @morgante I'm not sure if I'm following. Are you talking about the upgrade guide? 

On the provider we are just touching the default SAs.

bharathkkb:
@thiagonache I tested out a simple upgrade and I did not need to directly make any changes to the statefile.
My diff with initial `skip_gcloud=true` is below and it also worked `false`. What error did you observe?
```hcl
  # module.project-factory.module.project-factory.google_project_default_service_accounts.default_service_accounts will be created
  + resource "google_project_default_service_accounts" "default_service_accounts" {
      + action           = "DISABLE"
      + id               = (known after apply)
      + project          = "project-id"
      + restore_policy   = "REVERT"
      + service_accounts = (known after apply)
    }

  # module.project-factory.module.project-factory.module.gcloud_disable.null_resource.run_command[0] will be destroyed
  - resource "null_resource" "run_command" {
      - id       = "5504134643976332785" -> null
      - triggers = {
          - "activated_apis"          = "compute.googleapis.com,container.googleapis.com"
          - "arguments"               = "5cba1c8ef7cb6cc82509abf4392b5896"
          - "create_cmd_body"         = <<~EOT
                --project_id='project-id' \
                --sa_id='num-compute@developer.gserviceaccount.com' \
                --credentials_path='' \
                --impersonate-service-account='' \
                --action='disable'
            EOT
          - "create_cmd_entrypoint"   = ".terraform/modules/project-factory/modules/core_project_factory/scripts/modify-service-account.sh"
          - "default_service_account" = "num-compute@developer.gserviceaccount.com"
          - "gcloud_bin_abs_path"     = "/google-cloud-sdk/bin"
          - "md5"                     = "8724d44955a417594c942e0101e4fe82"
          - "project_services"        = "project-id"
        } -> null
    }

  # module.project-factory.module.project-factory.module.gcloud_disable.null_resource.run_destroy_command[0] will be destroyed
  - resource "null_resource" "run_destroy_command" {
      - id       = "1499773560248397990" -> null
      - triggers = {
          - "activated_apis"          = "compute.googleapis.com,container.googleapis.com"
          - "default_service_account" = "num-compute@developer.gserviceaccount.com"
          - "destroy_cmd_body"        = "info"
          - "destroy_cmd_entrypoint"  = "gcloud"
          - "gcloud_bin_abs_path"     = "/google-cloud-sdk/bin"
          - "project_services"        = "project-id"
        } -> null
    }
```

thiagonache:
It works, indeed.
After investigating I've found that the symlink I've created for the beta provider was pointing to the nonbeta provider :\
So I think we should add a note to upgrade the default_service_account to upper case if it is set somewhere.
Thank you for support.

thiagonache:
> > > @thiagonache I have some concerns about this deleting all SAs which have the role on existing projects. IMO we should actually filter which SAs are deleted (as we do in the script).
> > 
> > 
> > @morgante I'm not sure if I'm following. Are you talking about the upgrade guide?
> 
> On the provider we are just touching the default SAs.

Maybe you're talking about passing the service account id. My understanding is since the documentation says it is only for two APIs (compute engine and app engine) and the e-mail may change, the best way to handle it is directly in the provider. Also, to include a new service account is just about three lines of code in the provider. I'm not anticipating that changing, are you?

thiagonache:
@morgante Please, let me know if we are in agreement or if you want to change this.

thiagonache:
@bharathkkb do you have any consideration/opinion about the discussion above?

thiagonache:
@bharathkkb @morgante the new version is released. Can you guys run the jobs again to see if it passes, please?

thiagonache:
the lint error does not seem to be related to the change

morgante:
@thiagonache I think it might be actually:

```
terraform_validate ./test/fixtures/full 

No provider "google" plugins meet the constraint " < 4.0, < 4.0,3.40.0,>= 2.1,>= 3.47,~> 3.30".

The version constraint is derived from the "version" argument within the
provider "google" block in configuration. Child modules may also apply
provider version constraints. To view the provider versions requested by each
module in the current configuration, run "terraform providers".

To proceed, the version constraints for this provider must be relaxed by
either adjusting or removing the "version" argument in the provider blocks
throughout the configuration.
```

thiagonache:
> @thiagonache I think it might be actually:
> 
> ```
> terraform_validate ./test/fixtures/full 
> 
> No provider "google" plugins meet the constraint " < 4.0, < 4.0,3.40.0,>= 2.1,>= 3.47,~> 3.30".
> 
> The version constraint is derived from the "version" argument within the
> provider "google" block in configuration. Child modules may also apply
> provider version constraints. To view the provider versions requested by each
> module in the current configuration, run "terraform providers".
> 
> To proceed, the version constraints for this provider must be relaxed by
> either adjusting or removing the "version" argument in the provider blocks
> throughout the configuration.
> ```

odd I got a bunch of
```
Provider "gsuite" not available for installation.

A provider named "gsuite" could not be found in the Terraform Registry.

This may result from mistyping the provider name, or the given provider may
be a third-party provider that cannot be installed automatically.

In the latter case, the plugin must be installed manually by locating and
downloading a suitable distribution package and placing the plugin's executable
file in the following directory:
    terraform.d/plugins/linux_amd64

Terraform detects necessary plugins by inspecting the configuration and state.
To view the provider versions requested by each module, run
"terraform providers".
```
Checking

thiagonache:
@morgante sir, I've fixed my issue but we still have two errors:
```
Checking for trailing whitespace
./helpers/setup-sa.sh:226:  
Error: Trailing whitespace found in the lines above.
```
It was added by you recently, wasn't?

and

```
terraform_validate ./test/fixtures/full 

Error: Unsupported argument

  on .terraform/modules/vpc/main.tf line 58, in resource "google_compute_subnetwork" "subnetwork":
  58:   enable_flow_logs         = lookup(var.subnets[count.index], "subnet_flow_logs", "false")
```

morgante:
Yes, the whitespace issue was my mistake and I can fix.

The test fixture issue is likely due to your upgrading the provider version, so you'll possibly need to update the network module version to remove the `enable_flow_logs` variable.

bharathkkb:
I believe network module was already fixed in the 2.0 release. @thiagonache can you try bumping it to the latest?

thiagonache:
> I believe network module was already fixed in the 2.0 release. @thiagonache can you try bumping it to the latest?

you the man!

thiagonache:
@morgante @bharathkkb everything looks good on my side now. One thing I'm still not sure is if you guys agree with the approach of finding default service accounts.

README.md
```diff
@@ -122,7 +122,8 @@ determining that location is as follows:
 | budget\_amount | The amount to use for a budget alert | `number` | `null` | no |
 | budget\_monitoring\_notification\_channels | A list of monitoring notification channels in the form `[projects/{project_id}/notificationChannels/{channel_id}]`. A maximum of 5 channels are allowed. | `list(string)` | `[]` | no |
 | credentials\_path | Path to a service account credentials file with rights to run the Project Factory. If this file is absent Terraform will fall back to Application Default Credentials. | `string` | `""` | no |
-| default\_service\_account | Project default service account setting: can be one of `delete`, `deprivilege`, `disable`, or `keep`. | `string` | `"disable"` | no |
+| default\_sa\_restore\_policy | The action to be performed in the default service accounts on the resource destroy. Valid values are NONE and REVERT. If set to REVERT it will attempt to restore all default SAs but in the DEPRIVILEGE action. | `string` | `"REVERT"` | no |
```
morgante, line 125: I don't think we need to make this configurable.

morgante (changes requested):

README.md
thiagonache: it must be... if we DELETE it won't be able to revert after 30 days. So, the user must have to NONE.

thiagonache:
@morgante I've requested a review again since we cannot apply the last change requested.

README.md
morgante: It's fine for the behavior to be customizable on the provider, with the caveat that if set to "REVERT" I think the failure should not be an error (just a warning). /cc @rileykarson

rileykarson: I disagree, I think. Terraform failing to do what you told it to (eg undeleting the deleted SAs) should result in an error. We could consider an additional option like "REVERT_AND_IGNORE_FAILURE" though, I guess.

thiagonache: I think we can add another ENUM on_failure IGNORE or FAIL, but it would be an enhancement. I can work on it, but we should implement as-is for now.

morgante: Normally, I agree. But I'm not sure an error which users can't resolve in any way is particularly helpful.

I'd be fine if we wanted to split it to an additional `REVERT_AND_IGNORE_FAILURE` option instead though.

@thiagonache Can you work on that? For this PR we can stick to `REVERT` and remove the configurability.

thiagonache: > Normally, I agree. But I'm not sure an error which users can't resolve in any way is particularly helpful.
> 
> I'd be fine if we wanted to split it to an additional `REVERT_AND_IGNORE_FAILURE` option instead though.
> 
> @thiagonache Can you work on that? For this PR we can stick to `REVERT` and remove the configurability.

I can for sure, but by making it configurable the user can take an action to change to NONE, no?
Also, any strong preference using REVERT_AND_IGNORE_FAILURE vs another ENUM on_failure with values IGNORE or FAIL? I prefer the second one REVERT_AND_IGNORE_FAILURE looks odd, but maybe it's just me being annoying.

rileykarson: REVERT_AND_IGNORE_FAILURE imo- "NONE" and "IGNORE" doesn't make much sense to configure.

morgante: I explicitly don't want this configurable. It's not configurable today and we don't shouldn't change the surface area of the module as part of this change.

I'd prefer `REVERT_AND_IGNORE_FAILURE` as only this specific failure would be ignored (we would *not* ignore failures on original deletion, for example).

thiagonache: If it is not going to be configurable, why do we want to have a variable?

morgante: We don't. I want you to remove the variable.

thiagonache: in the provider too? I mean in the provider

thiagonache: we don't need to have a restore_policy variable there.

morgante: There are two different things which need to happen:

1. On the provider, update the `restore_policy` param to accept a `REVERT_AND_IGNORE_FAILURE` option.
2. On this module, remove the variable for restore policy and hardcode to `REVERT_AND_IGNORE_FAILURE`.

thiagonache: it does not make sense for me, but okay.

thiagonache: you mean it should accept REVERT and REVERT_AND_IGNORE_FAILURE on the provider?

thiagonache: yeah... it's clear now.

thiagonache:
so, can we merge it?

README.md
morgante: Is it still unclear? Please remove the variable.

thiagonache: I thought we are going to merge like that and just remove the variable and the provider handles the error... to me, it does not make sense to remove if it may cause an unrecoverable issue... but it's your call :)

morgante: We can set to `DELETE` for now then and change to `REVERT_AND_IGNORE_FAILURE` once it's available in the module. I want to avoid introducing a variable we will remove soon after.

thiagonache: you mean REVERT, right?
the problem I see is if we delete the service account now and we have no variable the new version **must** be released in less than 30 days.

morgante: I'm assuming you can work on a provider fix fairly quickly? We could also hold back this release until that's in.

thiagonache: which I think is feasible unless something really bad happens.

thiagonache: I'll be able to work on it next week and should take no more than one day. So, yeah, it's not a biggy to merge like that... we will have the new version before 30 days for sure

morgante: Sounds good!

docs/upgrading_to_project_factory_v10.0.md
```diff
@@ -0,0 +1,193 @@
+# Upgrading to Project Factory v10.0
+
+The v10.0 release of Project Factory is a backwards incompatible release for
+all modules since the breaking change is on
+[core_project_factory](../modules/core_project_factory) module which removes the
+need of gcloud and local-execs.
+
+## Migration Instructions
+
+1. Update variable default_service_account from down case to upper case if
```
morgante, line 10: We should not force users to change this. There's no reason we can't handle it easily in the module for them: https://www.terraform.io/docs/configuration/functions/upper.html

docs/upgrading_to_project_factory_v10.0.md
```diff
@@ -0,0 +1,193 @@
+# Upgrading to Project Factory v10.0
```
morgante, line 1: We also need to note that users will need to remove any references to `skip_gcloud_download` and `use\_tf\_google\_credentials\_env\_var`.

examples/simple_project/main.tf
```diff
@@ -46,7 +46,7 @@ module "project-factory" {
   org_id                  = var.organization_id
   billing_account         = var.billing_account
   credentials_path        = local.credentials_file_path
-  default_service_account = "deprivilege"
+  default_service_account = "DEPRIVILEGE"
```
morgante, line 49: Leave this unchanged to test the `upper()` functionality.

morgante (changes requested):
Almost there.

docs/upgrading_to_project_factory_v10.0.md
thiagonache: i know the upper function... since it's a breaking change I just thought a cleaner way would be better. i was wrong again :)

docs/upgrading_to_project_factory_v10.0.md
thiagonache: good catch.

README.md
thiagonache: https://github.com/hashicorp/terraform-provider-google/pull/7766 I still need to work on tests but we can start to discuss it

docs/upgrading_to_project_factory_v10.0.md
```diff
@@ -0,0 +1,190 @@
...
+
+The v10.0 release of Project Factory is a backwards incompatible release for
+all modules since the breaking change is on
+[core_project_factory](../modules/core_project_factory) module which removes the
+need of gcloud and local-execs.
+
+## Migration Instructions
+
+1. Remove any references of `skip_gcloud_download and use_tf_google_credentials_env_var` if applicable.
```
morgante, line 10:
```suggestion
Remove any references to `skip_gcloud_download and use_tf_google_credentials_env_var` if applicable.
```

morgante (approved):

PR #493, opened 2020-11-05
fix: Add billingbudgets.googleapis.com to precondition script.

### LINKED ISSUE #486: seed project script misses dependency of billing budget feature
```
"billingbudgets.googleapis.com"
```

Should be enabled for the seed project, to avoid an error when provisioning the first project / host-project / service-project.

morgante:
Fixes #486.


thiagonache:
should we also add check in the preconditions script?

morgante:
@thiagonache No I don't think a check is necessary. The error from this is fairly clear and recoverable.

bharathkkb (approved):
LGTM
CI failure seems unrelated

PR #500, opened 2020-11-17
fix: support passing service project number to shared_vpc_access to be tf 0.13 compatible

### LINKED ISSUE #499: shared_vpc module broken in 0.13
Sample config:

```
terraform {
  required_version = ">=0.12, <0.14"
  required_providers {
    google      = "~> 3.0"
    google-beta = "~> 3.0"
  }
}

module "project" {
  source  = "terraform-google-modules/project-factory/google//modules/shared_vpc"
  version = "~> 9.2.0"

  name                    = "example-data"
  org_id                  = ""
  folder_id               = "12345678"
  billing_account         = "000-000-000"
  lien                    = true
  default_service_account = "keep"
  skip_gcloud_download    = true

  shared_vpc = "example-networks"
  activate_apis = [
    "bigquery.googleapis.com",
    "compute.googleapis.com",
    "healthcare.googleapis.com",
    "servicenetworking.googleapis.com",
    "sqladmin.googleapis.com",
  ]
}
```

Running plan or apply results in the following failure:
```
Error: the user does not have permission to access Project "example-data" or it may not exist
```

The culprit is https://github.com/terraform-google-modules/terraform-google-project-factory/blob/a3deaad77881a09e73104984124cdd0740a09e8f/modules/shared_vpc_access/main.tf#L17 caused by https://github.com/hashicorp/terraform/issues/26857.

This config will not work because the project has not yet been created, but it is doing a lookup via the data source.

Solution here would be to make the shared_vpc_access module explicitly depend on the core project factory module (tf 0.13 only), or explicitly pass dependencies down to the shared vpc access module (tf 0.12 compatible).

The terraform team also recommends NOT using data sources in the same deployment as the resource creation, so a better solution here would be to pass the project number explicitly as a var.

cc @bharathkkb

umairidris:
Fix #499

### DIFF
--- a/modules/shared_vpc_access/README.md
+++ b/modules/shared_vpc_access/README.md
@@ -30,8 +30,9 @@ module "shared_vpc_access" {
 | active\_apis | The list of active apis on the service project. If api is not active this module will not try to activate it | `list(string)` | `[]` | no |
 | grant\_services\_security\_admin\_role | Whether or not to grant Kubernetes Engine Service Agent the Security Admin role on the host project so it can manage firewall rules | `bool` | `false` | no |
 | host\_project\_id | The ID of the host project which hosts the shared VPC | `string` | n/a | yes |
-| module\_depends\_on | List of modules or resources this module depends on. | `list(any)` | `[]` | no |
+| lookup\_project\_number | Whether to look up the project number from a data source. If false, `service_project_number` will be used instead. | `bool` | `true` | no |
 | service\_project\_id | The ID of the service project | `string` | n/a | yes |
+| service\_project\_number | Project number of the service project. Will be used if `lookup_service_project_number` is false. | `string` | `null` | no |
 | shared\_vpc\_subnets | List of subnets fully qualified subnet IDs (ie. projects/$project\_id/regions/$region/subnetworks/$subnet\_id) | `list(string)` | `[]` | no |
 
 ## Outputs
--- a/modules/shared_vpc_access/README.md
+++ b/modules/shared_vpc_access/README.md
@@ -30,7 +30,7 @@ module "shared_vpc_access" {
 | active\_apis | The list of active apis on the service project. If api is not active this module will not try to activate it | `list(string)` | `[]` | no |
 | grant\_services\_security\_admin\_role | Whether or not to grant Kubernetes Engine Service Agent the Security Admin role on the host project so it can manage firewall rules | `bool` | `false` | no |
 | host\_project\_id | The ID of the host project which hosts the shared VPC | `string` | n/a | yes |
-| lookup\_project\_number | Whether to look up the project number from a data source. If false, `service_project_number` will be used instead. | `bool` | `true` | no |
+| lookup\_project\_numbers | Whether to look up the project numbers from data sources. If false, `service_project_number` will be used instead. | `bool` | `true` | no |
 | service\_project\_id | The ID of the service project | `string` | n/a | yes |
 | service\_project\_number | Project number of the service project. Will be used if `lookup_service_project_number` is false. | `string` | `null` | no |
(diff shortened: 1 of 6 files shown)


morgante:
@umairidris Would it be possible to use implicitly dependency to solve this instead? I'd rather not be adding more `module_depends_on` resources now that Terraform 0.13 supports it natively.

bharathkkb:
We output `project_number` as part of core, so if we take that as an input for `shared_vpc_access` wouldn't that get rid of the need for module_depends_on?

umairidris:
@morgante implicit dependencies don't work as intended due to https://github.com/hashicorp/terraform/issues/26857.

@bharathkkb yes that was one option, but we will need to add a required var for project number which will be a breaking change on shared_vpc_access. Is that ok?

morgante:
Wouldn't an optional variable be sufficient?

umairidris:
Yes, that would work too though we would have two different paths (one with data source and one without). I can do that for now and we can remove the data source in a future major version.

umairidris:
Or not...

```
Error: Invalid count argument

  on ../../modules/shared_vpc_access/main.tf line 18, in data "google_project" "service_project":
  18:   count      = var.service_project_number == null ? 1 : 0

The "count" value depends on resource attributes that cannot be determined
resources that the count depends on.
```

bharathkkb:
I think it's okay to have a breaking change with the upcoming v10.0 for `shared_vpc_access`. It is mostly used as a helper for `shared_vpc` module but I will defer to @morgante

morgante:
This submodule is actually used directly frequently as well, so I'd definitely like to avoid a breaking change.

How about we make a `lookup_project_number` boolean variable as well (default `true`)?

umairidris:
If a v10 release is coming soon I would prefer to remove the data source entirely. It is more idiomatic it seems to put the usage of data and provider blocks in the user's control. In the meantime I have added the extra var.

morgante:
I'm strongly against any change which forces users to provide their project number. This module is scattered all over and, frankly, is simply shifting the burden of confusion to the user's config and reducing the value of the module.

I also think this error is one we're unlikely to encounter outside project factory, so I'd rather take on a bit more complexity here in exchange for easier interfaces outside the module.

umairidris:
That's a fair point, chances of this happening outside of PF is lower.

modules/shared_vpc_access/README.md
```diff
@@ -30,7 +30,9 @@ module "shared_vpc_access" {
 | active\_apis | The list of active apis on the service project. If api is not active this module will not try to activate it | `list(string)` | `[]` | no |
 | grant\_services\_security\_admin\_role | Whether or not to grant Kubernetes Engine Service Agent the Security Admin role on the host project so it can manage firewall rules | `bool` | `false` | no |
 | host\_project\_id | The ID of the host project which hosts the shared VPC | `string` | n/a | yes |
+| lookup\_project\_numbers | Whether to look up the project numbers from data sources. If false, `service_project_number` will be used instead. | `bool` | `true` | no |
```
bharathkkb, line 33: I saw that you made this plural - `project numbers from data sources`. Aren't we just using a single data source to look up a single project ?

bharathkkb (approved):
LGTM
/cc @morgante

modules/shared_vpc_access/README.md
morgante: I assume this is defensive naming for a future case where we might have to look up the host project number. Seems reasonable.

umairidris: Right, before it was a bit (lookup_service_project_number). But given @morgante comment above about continuing to support data sources, I imagine we may need to reuse this pattern in other places if this issue comes up, so using a generic/reusable var name will be more consistent for future uses.

bharathkkb: gotcha, thanks for clarifying.

umairidris:
@morgante does this look good to you?

morgante (approved):

PR #503, opened 2020-11-25
Increase project prefix length

jaroslawgajewski:
During extensive testing it popup that project IDs are randomly overlapping with existing projects when same names were used in different parent folders.  Problem mostly impacting development setups and infrastructures responsible for automated testing.  As result I'd like to propose to increase prefix length from 4 to 8 signs by increase of byte_length from 2 to 4

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -18,7 +18,7 @@
   Project random id suffix configuration
  *****************************************/
 resource "random_id" "random_project_id_suffix" {
-  byte_length = 2
+  byte_length = 4
 }
 
 /******************************************


morgante:
Unfortunately this would be a backwards-incompatible change that broke all existing project IDs.

I'm a little surprised that you're encountering issues here—we use this for all our CI projects and don't encounter collisions. Are you trying to name all projects the same?

jaroslawgajewski:
@morgante 
Yes we were using same names of projects placed under different folders within same organization. Till now issue popup once. 
in such case I'll have a look into frequency of failures to check if that was just bad luck 
thank you

morgante:
@jaroslawgajewski I'm going to close this for now, but if the issue continues to happen I'd be willing to review a pull request which makes the suffix length tunable.

PR #505, opened 2020-12-03
chore(terraform): support terraform 0.14 by bumping version constraint to v0.15

tomchinery:
This PR adds support for terraform 0.14.

This PR just bumps the version I haven't been able to test in my setup - I assume however it works as per https://github.com/terraform-google-modules/terraform-google-sql-db/pull/165#issue-531627730.

### DIFF
--- a/examples/app_engine/versions.tf
+++ b/examples/app_engine/versions.tf
@@ -15,5 +15,5 @@
  */
 
 terraform {
-  required_version = ">=0.12.6, <0.14"
+  required_version = ">=0.12.6, <0.15"
 }
--- a/examples/app_engine/versions.tf
+++ b/examples/app_engine/versions.tf
@@ -15,5 +15,5 @@
  */
 
 terraform {
-  required_version = ">=0.12.6, <0.15"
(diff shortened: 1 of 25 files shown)


bouk:
I think we should just remove the upper version constraint

tomchinery:
@bouk I agree. What stopped me was the other PR referenced above and that it already exists as a constraint. 

Terraform 0.13.X generates version constraints automatically for the minimum versions and not the maximum versions. It's best practice to only specify a minimum version constraint as per the docs: https://www.terraform.io/docs/configuration/version-constraints.html#terraform-core-and-provider-versions. 

Must be a reason a maximum version is specified here? If not happy to commit an additional commit to  remove completely.

bharathkkb:
@tomchinery feel free to remove the upper constraint

morgante:
The main reason we have an upper version constraint was due to earlier incompatibilities. We can certainly drop it now.

tomchinery:
Added additional commit removing the version constraint. Thanks all for feedback + quick response.

morgante (approved):

PR #513, opened 2020-12-09
feat!: add TF 0.13 constraint and module attribution

bharathkkb:
- add module attribution
- add TF 0.13 constraint

### DIFF
--- a/modules/core_project_factory/versions.tf
+++ b/modules/core_project_factory/versions.tf
@@ -35,10 +35,4 @@ terraform {
       version = "~> 2.2"
     }
   }
-  provider_meta "google" {
-    module_name = "blueprints/terraform/terraform-google-project-factory:core_project_factory/v10.0.0"
-  }
-  provider_meta "google-beta" {
-    module_name = "blueprints/terraform/terraform-google-project-factory:core_project_factory/v10.0.0"
-  }
 }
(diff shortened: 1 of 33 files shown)


morgante:
@bharathkkb Looks like tests are failing.

bharathkkb:
@morgante It was failing due to the 3rd party provider. I have updated to use the new source syntax for both 3rd party and google providers.

modules/core_project_factory/versions.tf
```diff
@@ -15,12 +15,30 @@
...
+      source  = "hashicorp/null"
+      version = "~> 2.1"
+    }
+    random = {
+      source  = "hashicorp/random"
+      version = "~> 2.2"
+    }
+  }
+  provider_meta "google" {
+    module_name = "blueprints/terraform/terraform-google-project-factory:core_project_factory/v10.0.0"
```
morgante, line 39: We shouldn't use add the `provider_meta` for internal modules so let's drop this section.

morgante (changes requested):

modules/core_project_factory/versions.tf
bharathkkb: sg; in that case should `shared_vpc` be an internal module or should we add a readme for it?

bharathkkb: done

morgante (approved):

hrvolapeter:
@bharathkkb was there a reason for requiring google providers less than 4.0.0 ? 4.0.0 is out and this is preventing from upgrading to the latest version.

Issue #519, opened 2020-12-16
Using PF 10.0.0 with TF 0.13.5 without random_project_id failed


xingao267:
Example to reproduce:

```
module "project" {
  source  = "terraform-google-modules/project-factory/google"
  version = "~> 10.0.0"

  name                    = "example-project"
  org_id                  = ""
  folder_id               = "12345678"
  billing_account         = "000-000-000"
  lien                    = true
  default_service_account = "keep"
  activate_apis = [
    "cloudbuild.googleapis.com",
  ]
}
```

Error when running `terraform plan`:
```
$ terraform plan
Refreshing Terraform state in-memory prior to plan...
The refreshed state will be used to calculate this plan, but will not be
persisted to local or remote state storage.


------------------------------------------------------------------------

Error: the user does not have permission to access Project "example-project" or it may not exist
```

morgante:
If you use a more unique project ID, does it work?

xingao267:
Changed it to `example-devops-xingao-20201216` and still gives the same error:

```
Error: the user does not have permission to access Project "example-devops-xingao-20201216" or it may not exist
```

morgante:
Just to confirm, is this a totally fresh Terraform state? Does the error log clarify which resource it is failing on?

I'm not sure why the random ID part would cause a difference here.

xingao267:
Yes, it is a totally fresh Terraform state. I did more exploration and tests. `svpc_service_project` is fine with TF 0.13. My guess is https://github.com/terraform-google-modules/terraform-google-project-factory/commit/825d07b19417827d6ad66f9a8dd437b53de32bbc should also be applied to the root module as `shared_vpc_access` is also added to the module as in https://github.com/terraform-google-modules/terraform-google-project-factory/pull/446/files. Let me verify this and send a PR.

PR #520, opened 2020-12-16
fix: pass service project num to shared_vpc_access in root module

xingao267:
Fixes https://github.com/terraform-google-modules/terraform-google-project-factory/issues/519


morgante (approved):

morgante:
@xingao267 Do we need this for the `gsuite` submodule as well?

xingao267:
@morgante i don't think so. Only in modules where `shared_vpc_access` submodule is called.

morgante:
Looks like we don't have the submodule there at all, opened #522 to track.

PR #528, opened 2021-01-06
Closes #526: Service account creation optional

### LINKED ISSUE #526: Feature Request: Service Account creation optional
Hey everyone,

currently the default service account for terraform always gets created (see: [here](https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/core_project_factory/main.tf#L131)). We don't really need this service account and our security team is arguing that we should not have this service account created, even it is has no rights associated with it. Since we don't want to fork this and disable the service account deletion on our own, we would love to get this in here.

I'd be very happy with issuing a Merge request here that introduces an optional flag to skip the service account creation which is per default set to `false`. 

Any options on this?

sebastianneb:
This MR resolves #526. I added the flag `create_project_sa` and set it to `true` as default which makes this change an Addition and not a Breaking change.
I also adopted the Contributing guide with new descriptions over which I stumbled while executing the tests locally.
Apart from that I did not add a test for the new flag. I can surely create a new Inspec control for this if you wish.

Any feedback appreciated 👍


morgante (approved):
Thanks for adding the CONTRIBUTING.md changes too.

PR #531, opened 2021-01-13
feat: add support for multiple roles to assign project service account

cleibl:
This PR allows for the ability to bind more than one role to the created project service account

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -138,9 +138,9 @@ resource "google_service_account" "default_service_account" {
   Policy to operate instances in shared subnetwork
  *************************************************/
 resource "google_project_iam_member" "default_service_account_membership" {
-  count   = var.sa_role != "" ? 1 : 0
+  for_each = toset(var.sa_roles)
   project = google_project.main.project_id
-  role    = var.sa_role
+  role    = each.key
 
   member = local.s_account_fmt
 }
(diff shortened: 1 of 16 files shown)


modules/core_project_factory/main.tf
```diff
@@ -139,9 +139,9 @@ resource "google_service_account" "default_service_account" {
   Policy to operate instances in shared subnetwork
  *************************************************/
 resource "google_project_iam_member" "default_service_account_membership" {
-  count   = var.sa_role != "" && var.create_project_sa ? 1 : 0
-  project = google_project.main.project_id
-  role    = var.sa_role
+  for_each = toset(var.sa_roles)
```
morgante, line 142: This should also still test for `var.create_project_sa`.

variables.tf
```diff
@@ -77,18 +77,18 @@ variable "group_role" {
...
+  default     = []
+}
+
 variable "create_project_sa" {
   description = "Whether the default service account for the project shall be created"
   type        = bool
   default     = true
 }
 
-variable "sa_role" {
```
morgante, line 86: To preserve backwards-compatibility, let's continue accepting `sa_role` on the "outer" modules but add a note to the description that it is deprecated.

morgante (changes requested):
Thanks for the contribution! 💯

PR #534, opened 2021-01-14
feat: Add labels support to projects bucket

akovtun-extenda:
This PR adds support for adding labels to the bucket


bharathkkb (approved):
Thanks for the PR @akovtun-extenda

PR #550, opened 2021-02-23
fix: Change budget module: Accept list of project numbers instead of …

### LINKED ISSUE #544: Change budget module: Accept list of project numbers instead of project ID
[google_billing_budget](https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/billing_budget) now accepts projects as projects/{project_number}. 

projects - (Optional) A set of projects of the form **projects/{project_number}**, specifying....

Can we change [budget module](https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/budget) to accept list of project numbers instead of project IDs.

Also make following change in main module:

```
module "budget" {
  source        = "./modules/budget"
  create_budget = var.budget_amount != null

  projects                         = [module.project-factory.project_number]

```

switj:
…project ID #544

Hello 👋  this is my first PR. 

I've reviewed the contributing guidelines but if I've missed anything please direct me at the relevant doc/section so I can fix the issue.

Fixes #544

### DIFF
--- a/modules/budget/main.tf
+++ b/modules/budget/main.tf
@@ -17,16 +17,21 @@
 locals {
   project_name     = length(var.projects) == 0 ? "All Projects" : var.projects[0]
   display_name     = var.display_name == null ? "Budget For ${local.project_name}" : var.display_name
-  all_updates_rule = var.alert_pubsub_topic == null && var.monitoring_notification_channels == null ? [] : ["1"]
+  all_updates_rule = var.alert_pubsub_topic == null && length(var.monitoring_notification_channels) == 0 ? [] : ["1"]
 
   projects = length(var.projects) == 0 ? null : [
-    for id in var.projects :
-    "projects/${id}"
+    for project in data.google_project.project :
+    "projects/${project.number}"
   ]
   services = var.services == null ? null : [
     for id in var.services :
     "services/${id}"
   ]
+}
+
+data "google_project" "project" {
+  count      = length(var.projects)
+  project_id = element(var.projects, count.index)
 }
 
 resource "google_billing_budget" "budget" {


comment-bot-dev:

morgante (changes requested):
Unfortunately this causes an error:
```
       
         on ../../../modules/budget/main.tf line 33, in data "google_project" "project":
         33:   for_each   = toset(var.projects)
       
       The "for_each" value depends on resource attributes that cannot be determined
       resources that the for_each depends on.
```

modules/budget/main.tf
```diff
@@ -17,18 +17,23 @@
 locals {
   project_name     = length(var.projects) == 0 ? "All Projects" : var.projects[0]
   display_name     = var.display_name == null ? "Budget For ${local.project_name}" : var.display_name
-  all_updates_rule = var.alert_pubsub_topic == null && var.monitoring_notification_channels == null ? [] : ["1"]
+  all_updates_rule = var.alert_pubsub_topic == null && length(var.monitoring_notification_channels) == 0 ? [] : ["1"]
 
   projects = length(var.projects) == 0 ? null : [
-    for id in var.projects :
-    "projects/${id}"
+    for project in data.google_project.project :
```
morgante, line 23: What I think you're going to need to do is:

Keep the `for_each` the same (iterating over the project_id).

Inside the resource itself, look up the number from the data source.

switj: I was able to repro using the integration tests. Not all tests are working (I don't think the errors are related to my changes) but the billing test seems ok now.

Switching from `for_each` to `count` seems to solve the issue.

switj:
@morgante thanks for identifying the error and sharing so I could address the issue.

I had some issues setting up the integration tests which I've since resolved. There are still a few tests that don't seem to run properly with my org but overall I think I'm in a better place to catch these issues myself.

modules/budget/main.tf
morgante: Thanks, looks like that worked.

morgante:
Thanks @switj, this looks good and I think the test failure is ephemeral.

switj:
hiya 👋 @morgante bumping this PR. 

I'm happy to rework the PR or spend additional time looking at ephemeral test failures if that makes the review easier. If there is anything I can help let me know.

morgante (approved):
I think the error is unrelated.

PR #567, opened 2021-03-24
fix: Composer shared vpc

diegolnasc:
Composer needs role _composer.sharedVpcAgent_ or _networkAdmin_ at the project level because of peering. Therefore, this hotfix separetes the role in another resource in the same way that we are implementing it in the gke agent.


comment-bot-dev:

morgante (approved):

PR #569, opened 2021-03-30
Change Quota Manager Terraform Required Version

lawliet89:
The root module support `>= 0.13.0`. There is no reason for one single module to break for existing users on Terraform 0.14, especially with the new release.

I have set it to be consistent with the root module.

### DIFF
--- a/modules/quota_manager/versions.tf
+++ b/modules/quota_manager/versions.tf
@@ -15,7 +15,7 @@
  */
 
 terraform {
-  required_version = ">=0.12.6, <0.14"
+  required_version = ">=0.13.0"
 
   required_providers {
     google-beta = ">= 3.1, < 4.0"


comment-bot-dev:

ToxicWar:
@lawliet89 could we keep `>=0.12.6` as in all other modules?

https://github.com/terraform-google-modules/terraform-google-project-factory/pull/505/files

modules/quota_manager/versions.tf
```diff
@@ -15,7 +15,7 @@
  */
 
 terraform {
-  required_version = ">=0.12.6, <0.14"
```
ToxicWar, line 18: @lawliet89 could we keep >=0.12.6 as in all other modules?

https://github.com/terraform-google-modules/terraform-google-project-factory/pull/505/files

morgante: We've switched to requiring 0.13.

morgante (approved):

PR #432, opened 2020-07-20
fix: Add dependency on svpc

bharathkkb:
Wait for `google_compute_shared_vpc_service_project` to resolve before output `project_id`

PR #572, opened 2021-03-31
Simplify project_id output in core_project_factory

MartinPetkov:
I ran into a case where somehow `google_compute_shared_vpc_service_project.shared_vpc_attachment` was an "empty tuple" in the state, and this output was failing because it was trying to index into it. I came across https://github.com/hashicorp/terraform/issues/25578, which might be related, but it does still happen with the latest version of Terraform as of opening this bug (v0.14.9).

Regardless, it seems to me that all this is doing is building a list then taking the first element, so the one-liner should be equivalent and avoids this error.

### DIFF
--- a/modules/core_project_factory/outputs.tf
+++ b/modules/core_project_factory/outputs.tf
@@ -19,8 +19,13 @@ output "project_name" {
 }
 
 output "project_id" {
-  value      = module.project_services.project_id
-  depends_on = [module.project_services]
+  value = module.project_services.project_id
+  depends_on = [
+    module.project_services,
+    google_project.main,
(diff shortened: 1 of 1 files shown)


comment-bot-dev:

bharathkkb (commented):
IIRC when building the list it will resolve all elements and then take the first element so we are able to wait on svpc which was added in https://github.com/terraform-google-modules/terraform-google-project-factory/pull/432.  I don't think we should remove the wait on svpc, perhaps we can use `try` to catch the null error?

morgante (changes requested):
Let's make sure to keep the same dependencies list.

morgante (approved):

PR #574, opened 2021-04-09
Added region tags for shared VPC host and service projects

betsy-lichtenberg:
Added region tags for shared VPC host and service projects
For b/184973701
Snippets to be included on https://cloud.google.com/vpc/docs/provisioning-shared-vpc

### DIFF
--- a/examples/shared_vpc/main.tf
+++ b/examples/shared_vpc/main.tf
@@ -43,7 +43,8 @@ provider "random" {
  *****************************************/
 # [START vpc_shared_vpc_host_project_create]
 module "host-project" {
-  source                         = "../../"
+  source                         = "terraform-google-modules/terraform-google-project-factory/google//modules/shared_vpc_access"
+  version                        = "~> 10.3.1"
   random_project_id              = true
   name                           = var.host_project_name
   org_id                         = var.organization_id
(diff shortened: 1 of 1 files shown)


comment-bot-dev:

examples/shared_vpc/main.tf
```diff
@@ -41,15 +41,18 @@ provider "random" {
 /******************************************
   Host Project Creation
  *****************************************/
+# [START vpc_shared_vpc_host_project_create]
 module "host-project" {
-  source                         = "../../"
+  source                         = "terraform-google-modules/terraform-google-project-factory/google//modules/shared_vpc_access"
```
morgante, line 46: This isn't the right source. What are you trying to show with this example?

morgante (changes requested):

examples/shared_vpc/main.tf
betsy-lichtenberg: I'm trying to add some Terraform tabs to this page: https://cloud.google.com/vpc/docs/provisioning-shared-vpc#setting_up

morgante: Got it. This isn't the right example for that, as this actually *creates* the host project. Is that what you want to do?

If you just want to activate Shared VPC on an existing project, a new example would be needed.

betsy-lichtenberg: Got it. Thanks for the guidance. I'm closing this PR and starting over with a new example.

PR #601, opened 2021-06-30
fix: Using project_id output is not forcing to wait for the project creation

### LINKED ISSUE #598: Using project_id output is not forcing to wait for the project creation
Tested on `v0.13.7`, `v0.15.3`

I am using `fabric_project` submodule in a following way: 

```
terraform {
  required_version = ">= 0.13"

  required_providers {
    google      = ">= 3.71.0"
  }
}

provider "google" {
  region = "europe-central2"
}

module "fabric-project" {
  source  = "terraform-google-modules/project-factory/google//modules/fabric-project"
  version = "11.0.0"

  name            = "test-project-name-43413"
  billing_account = "<BILLING_ACCOUNT_ID>"
  parent          = "folders/<FOLDER_ID>"
  prefix          = "dev"

  auto_create_network = false

  activate_apis = [
    "compute.googleapis.com",
  ]
}

data "google_storage_project_service_account" "gcs_account" {
  project = module.fabric-project.project_id
}
```

Because `google_storage_project_service_account` datasource is depending on `project_id` from `fabric-project`, it should wait for the project creation. But running `terraform plan` is actually running datasource's GET straight away:
```
Error: Error when reading or editing GCS service account not found: googleapi: Error 400: Unknown project id: 'dev-test-project-name-43413', invalid

  on main.tf line 29, in data "google_storage_project_service_account" "gcs_account":
  29: data "google_storage_project_service_account" "gcs_account" {
```

This is because `module.fabric_project.project_id` output is using `google_project.project.project_id` attribute which is not a computed attribute, so it is available for terraform immediately, so dependencies are immediately satisfied and this is causing that other resources using `project_id` do not wait for the project creation.  

This is output definition from `fabric-project/outputs.tf`
```
output "project_id" {
  description = "Project id (depends on services)."
  value       = google_project.project.project_id
  depends_on  = [google_project_service.project_services]
}
```

Despite declared `depends_on  = [google_project_service.project_services]` the `project_id` value is available immediately. I am not sure if this is a bug in a terraform, definitely it is not really an intuitive behavior. 

Anyway the fix is trivial, just use for `project_id` output value from `id` which is in the form of `projects/<project_id>` and that forces terraform to wait for a project resource.  Tell me if it makes sense - I have a pull request ready for it. Or maybe I am missing something?
 
```
output "project_id" {
  description = "Project id (depends on services)."
  value       = trimprefix(google_project.project.id, "projects/")
  depends_on  = [google_project_service.project_services, google_project.project]
}
```

lsierant:
This fixes #598. 

Using `project_id` output with this fix will force terraform to wait on project creation and resolve dependencies properly. This is primarily fix for a case when there are datasources that are gathering any data using `project_id`. These datasources are run immediately and fail in plan phase, because `project_id` value output was available immediately, not after creating the project. 

One warning to consider though. After this change `project_id` output value will not be available until after apply. This causes terraform to error if `project_id` output is used anywhere in for_each clauses. Terraform cannot handle resources without knowing upfront how many resources there will be. To avoid that, it is needed to perform `terraform apply -target module.project` to create a project first, then second standard apply to create resources that are depending on it. 

I haven't run any tests, because I don't know whether there is any existing test env or every contributor should test on their own gcp org.

### DIFF
--- a/modules/fabric-project/outputs.tf
+++ b/modules/fabric-project/outputs.tf
@@ -16,7 +16,7 @@
 
 output "project_id" {
   description = "Project id (depends on services)."
-  value       = google_project.project.project_id
+  value       = trimprefix(google_project.project.id, "projects/")
   depends_on  = [google_project_service.project_services]
 }
 


comment-bot-dev:

modules/fabric-project/outputs.tf
```diff
@@ -16,7 +16,7 @@
 
 output "project_id" {
   description = "Project id (depends on services)."
-  value       = google_project.project.project_id
+  value       = trimprefix(google_project.project.id, "projects/")
```
bharathkkb, line 19: Doesn't depends_on = [google_project. project] work?

bharathkkb (commented):
Thanks for the PR @lsierant

bharathkkb:
Also 
> these datasources are run immediately and fail in plan phase

Could you show an example. Usually you can defer datasources to apply time using depends_on within datasources.

lsierant:
> Also
> 
> > these datasources are run immediately and fail in plan phase
> 
> Could you show an example. Usually you can defer datasources to apply time using depends_on within datasources.

Example code: 
```hcl
terraform {
  required_version = ">= 0.13"

  required_providers {
    google      = ">= 3.71.0"
  }
}

provider "google" {
  region = "europe-central2"
}

module "project" {
  source  = "terraform-google-modules/project-factory/google//modules/fabric-project"
  version = "11.0.0"

  name            = "test-project-name-43413"
  billing_account = "<BILLING_ACCOUNT_ID>"
  parent          = "folders/<FOLDER_ID>"
  prefix          = "dev"

  auto_create_network = false

  activate_apis = [
    "compute.googleapis.com",
  ]
}

data "google_storage_project_service_account" "gcs_account" {
  project = module.project.project_id
}

output "project-id" {
  value = module.project.project_id
}
```

That causes datasource to fail:
```
$ terraform plan

 Error: Error when reading or editing GCS service account not found: googleapi: Error 400: Unknown project id: 'dev-test-project-name-43413', invalid

   with data.google_storage_project_service_account.gcs_account,
   on main.tf line 29, in data "google_storage_project_service_account" "gcs_account":
   29: data "google_storage_project_service_account" "gcs_account" {
```

Of course explicitly depending on project solves this issue, but it's counter-intuitive. 
```hcl
data "google_storage_project_service_account" "gcs_account" {
  project = module.project.project_id
  depends_on = [module.project]
}
```

Especially if you look into `project_id` output in `modules/fabric-project/outputs.tf`, where there is explicit depends_on not even on project but on project services. Which makes sense, because you don't want to use project_id if it doesn't have APIs enabled. 
```output "project_id" {
  description = "Project id (depends on services)."
  value       = google_project.project.project_id
  depends_on  = [google_project_service.project_services]
}
```
But this depends_on in output does not work, because project_id is available immediately. I'm not sure if this is not a bug in terraform, anyway current behaviour is surprising. `project_id` is often used in many resources and specifying explicitly dependencies on module.project is possible, but tedious.

modules/fabric-project/outputs.tf
lsierant: I'm not sure which depends_on you're asking:
1. if you're asking about depends_on here in output then no, changing 
```hcl
  depends_on  = [google_project_service.project_services]
```
to 
```hcl
  depends_on  = [google_project.project]
```
does not work. Terraform ignores this explicit dependency when it has the output value available immediately. This is especially the case when output value comes from input attribute. 

2. If you're asking whether explicit depends on module.project works e.g. in a datasource that is using this output, then yes, it's working. But please see my comment in main thread. 
```hcl
data "google_storage_project_service_account" "gcs_account" {
  project = module.project.project_id
  depends_on = [module.project]
}
```

bharathkkb: I see, I assumed depends_on may block on `google_project.project` since some of the attribs are computed.

bharathkkb (approved):

PR #628, opened 2021-10-09
Handle compute.googleapis.com properly when passed in via activate_api_identities

rsinnet:
## Synopsis

Process roles for the compute.googleapis.com service identity separately from other service identities when passed via activate_api_identities.

## Details

This service identity cannot be created with the `google_project_service_identity` resource. Attempting to activate this service identity leads to an error response. It seems that this API call is not permitted despite lack of documentation. This can be seen by calling:

```sh
gcloud alpha services identity create --service=compute.googleapis.com --log-http
```

Calling this same command with, e.g., cloudbuild.googleapis.com does not result in an error.

It seems the service is activated by default, so instead just retrieve the email using `google_compute_default_service_account`. So get this email, create a combined map with this service identity (if present) and other service identities (if present), and then process this map as before, adding each role for each service identity.

### DIFF
--- a/modules/project_services/main.tf
+++ b/modules/project_services/main.tf
@@ -15,7 +15,8 @@
  */
 
 locals {
-  services = var.enable_apis ? toset(concat(var.activate_apis, [for i in var.activate_api_identities : i.api])) : toset([])
+  activate_compute_identity = 0 != length([for i in var.activate_api_identities : i if i.api == "compute.googleapis.com"])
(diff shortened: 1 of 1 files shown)


comment-bot-dev:

morgante:
Instead of raising an error/warning, should we simply filter this value out of the list automatically?

rsinnet:
> Instead of raising an error/warning, should we simply filter this value out of the list automatically?

That's an interesting suggestion. I like it because it defines the error out of existence and the outcome is what the user would probably expect. I would be in favor of that interface. Since I'm in the code at the moment, I'll go ahead and push a proposed implementation.

Issue #639, opened 2021-11-30
Allow configuring default network service tier for projects


jawnsy:
### TL;DR

This proposes the addition of a setting to configure the default [Network Service Tier](https://cloud.google.com/network-tiers) for projects.

### Terraform Resources

https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_project_default_network_tier

### Detailed design

Add a `network_tier` field for the project and configure it. The default value should be `PREMIUM`, which is the default on Google Cloud.

morgante:
This seems like a reasonable addition, we'd be happy to review a pull request adding it.

PR #640, opened 2021-11-30
feat: add network service tier setting

jawnsy:
This allows users to configure the project default Network Service
Tier.

Closes: #639


modules/core_project_factory/variables.tf
```diff
@@ -165,7 +165,7 @@ variable "bucket_versioning" {
 
 variable "bucket_labels" {
   description = " A map of key/value label pairs to assign to the bucket (optional)"
-  type        = map
+  type        = map(any)
```
jawnsy, line 168: terraform fmt seemed to add this

```
$ terraform version
Terraform v1.0.11
on linux_amd64
```

modules/core_project_factory/main.tf
```diff
@@ -350,3 +350,12 @@ resource "google_project_service" "enable_access_context_manager" {
   project = google_project.main.number
   service = "accesscontextmanager.googleapis.com"
 }
+
+/******************************************
+  Configure default Network Service Tier
+ *****************************************/
+resource "google_compute_project_default_network_tier" "default" {
+  count        = var.default_network_tier != "" ? 1 : 0
```
jawnsy, line 358: I tried to follow the pattern in the rest of this module, this shouldn't do anything if the setting isn't set, otherwise it just passes through whatever the user defines. This seems like a safe way to implement this module, to avoid interfering with any existing installs

comment-bot-dev:
Thanks for the PR! 🚀
Unfortunately it looks like some of our CI checks failed. See the [Contributing Guide](../blob/master/CONTRIBUTING.md) for details.
- ⚠️check_documentation
 The documentation needs to be regenerated. Please run `make generate_docs`. 
 ```bash 
Checking for documentation generation
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=.git' '--exclude=autogen' '--exclude=*.tfvars' /workspace/README.md /tmp/tmp.HgxX0vnVCo/generate_docs/workspace/README.md
117c117
< | bucket\_labels | A map of key/value label pairs to assign to the bucket (optional) | `map` | `{}` | no |
---
> | bucket\_labels | A map of key/value label pairs to assign to the bucket (optional) | `map(any)` | `{}` | no |
129a130
> | default\_network\_tier | Default Network Service Tier for resources created in this project. See https://cloud.google.com/network-tiers/docs/using-network-service-tiers and https://cloud.google.com/network-tiers. | `string` | `""` | no |
Error: Documentation generation has not been run, please run the
'make docker_generate_docs' command and commit the above changes.
 ```

modules/core_project_factory/main.tf
morgante: Thanks, sounds good to me!

modules/core_project_factory/variables.tf
morgante: Since this is a map of strings, can you change it to `map(string)`?

variables.tf
```diff
@@ -275,3 +275,9 @@ variable "consumer_quotas" {
   }))
   default = []
 }
+
+variable "default_network_tier" {
+  description = "Default Network Service Tier for resources created in this project. See https://cloud.google.com/network-tiers/docs/using-network-service-tiers and https://cloud.google.com/network-tiers."
```
morgante, line 280: Clarify that if unset this value isn't updated at all.

morgante (changes requested):
Please also propagate the variable to the shared_vpc and gsuite variants.

variables.tf
jawnsy: thanks, added the following text (bolded for emphasis here):

>Default Network Service Tier for resources created in this project. **If unset, the value will not be modified.** See https://cloud.google.com/network-tiers/docs/using-network-service-tiers and https://cloud.google.com/network-tiers.

modules/core_project_factory/variables.tf
jawnsy: makes sense, done!

morgante (approved):

PR #725, opened 2022-06-01
feat: add deploymentmanager.googleapis.com to supported apis

bestefreund:
This adds the deployment manager-api to the supported apis.

This is mandatory for using GKE in a shared vpc-subnet when the nodes are getting deployed.

Example-Output:
```
Google Compute Engine: Not all instances running in IGM after 14.50192001s.
Expected 1, running 0, transitioning 1. Current errors:
[PERMISSIONS_ERROR]: Instance '<GKE-NODE-POOL-ID>' creation failed:
Required 'compute.subnetworks.use' permission for 'projects/<NETWORK-HOST-PROJECT-ID>/regions/<SUBNET-REGION>/subnetworks/<SUBNET-ID>'
(when acting as '<SERVICE-PROJECT-NUMBER>@cloudservices.gserviceaccount.com').
```

### DIFF
--- a/modules/shared_vpc_access/main.tf
+++ b/modules/shared_vpc_access/main.tf
@@ -27,6 +27,7 @@ locals {
     "dataflow.googleapis.com" : format("service-%s@dataflow-service-producer-prod.iam.gserviceaccount.com", local.service_project_number),
     "composer.googleapis.com" : format("service-%s@cloudcomposer-accounts.iam.gserviceaccount.com", local.service_project_number)
     "vpcaccess.googleapis.com" : format("service-%s@gcp-sa-vpcaccess.iam.gserviceaccount.com", local.service_project_number)
+    "deploymentmanager.googleapis.com" : format("%s@cloudservices.gserviceaccount.com", local.service_project_number)
   }
   gke_shared_vpc_enabled      = contains(var.active_apis, "container.googleapis.com")
   composer_shared_vpc_enabled = contains(var.active_apis, "composer.googleapis.com")


comment-bot-dev:
@bestefreund

bestefreund:
@jeffmccune 
@Jberlinsky 
@jawnsy 

Please, let me know, what's wrong with the pipeline to fix it

bestefreund:
@aaron-lane 
@morgante 
@adrienthebo 

@patricklubach

I've tested the pipeline locally & it seems that something is wrong with the test that is not related to my change.

Please, let me know, what to do.

The logs are attached.

[kitchen.log](https://github.com/terraform-google-modules/terraform-google-project-factory/files/8822902/kitchen.log)
[vpc-sc-project-local.log](https://github.com/terraform-google-modules/terraform-google-project-factory/files/8822903/vpc-sc-project-local.log)

modules/shared_vpc_access/main.tf
```diff
@@ -27,6 +27,7 @@ locals {
     "dataflow.googleapis.com" : format("service-%s@dataflow-service-producer-prod.iam.gserviceaccount.com", local.service_project_number),
     "composer.googleapis.com" : format("service-%s@cloudcomposer-accounts.iam.gserviceaccount.com", local.service_project_number)
     "vpcaccess.googleapis.com" : format("service-%s@gcp-sa-vpcaccess.iam.gserviceaccount.com", local.service_project_number)
+    "deploymentmanager.googleapis.com" : format("%s@cloudservices.gserviceaccount.com", local.service_project_number)
```
bharathkkb, line 30: `project-num@cloudservices.gserviceaccount.com` is the API service agent which exists at project creation and not tied to `deploymentmanager.googleapis.com`. We grant this the necessary roles in core project factory. 
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/d309270a6c6ace6e2b5bc792073bfee9a07c365d/modules/core_project_factory/main.tf#L41-L44
https://github.com/terraform-google-modules/terraform-google-project-factory/blob/d309270a6c6ace6e2b5bc792073bfee9a07c365d/modules/core_project_factory/main.tf#L179-L191
Are you using a different module for project creation?

bestefreund: Hi @bharathkkb,

thanks for the fast reply and sorry for the confusion. 

I thought it's tied to `deploymentmanager.googleapis.com` because of this article:
https://cloud.google.com/deployment-manager/docs/access-control#access_control_for

```
Access control for Deployment Manager

To create other Google Cloud resources, Deployment Manager uses the credentials of the Google APIs Service Agent to authenticate to other APIs. The Google APIs Service Agent is designed specifically to run internal Google processes on your behalf. This service account is identifiable using the email:

[PROJECT_NUMBER]@cloudservices.gserviceaccount.com
```

We are creating the service project in an environment other than the network host project containing the shared VPC.

The network host project exists before the service projects are deployed in our deployment chain & the shared_vpc_access is used in the same environment where the service projects are deplyoed.

For creating network host projects and service projects we use the project-factory module:

```
  source                  = "terraform-google-modules/project-factory/google"
  version                 = "~> 11.3"
```

To be honest, our DEV-teams are enabled to manage the shared_vpc_iam by their own & it could be that they have overwritten the networkUser-role.

Tomorrow I'll try to create a clean environment to check if the missed IAM-permissions are in place.

Thanks again!

Cheers!

bestefreund: Hi @bharathkkb,

One addtition:

We use the `shared_vpc_access`-module instead of providing the network host project info to the `project-factory`-module because in the past we had issues with it.

We used the following attributes to bind the created service projects to the existing network host projects:

```
 svpc_host_project_id
 shared_vpc_subnets
 grant_services_security_admin_role
```

In some cases Terraform stated replacements of `google_project`-resources from the `project-factory` when something has changed in our network host projects.

To be honest, it's long time ago when we made the switch from the three attributes from the `project-factory` to the `shared_vpc_access`-module and I cannot find the related Terraform-plan-output to prove it and to figure out what in detail triggered the replacements.

Nevertheless is it possible in general to include the API service agent into the `shared_vpc_access`-module?

If yes, let me know which API should be provided.

Another option could be that I add a new module-attribute to assign the permissions to the API service agent, if there is no related API to the API service agent.

If you have another suggestion, just let me know.

If nothing helps, please, just decline my PR and we'll assign the permissions without the `shared_vpc_module`.

Cheers!

bharathkkb: @bestefreund sorry for the late reply.
> In some cases Terraform stated replacements of google_project-resources from the project-factory when something has changed in our network host projects.

This seems like a bug, if you encounter this again feel free to open an issue.

>  Is it possible in general to include the API service agent into the shared_vpc_access-module

This seems reasonable although we may need to do some investigation if it can co exist without a flag for https://github.com/terraform-google-modules/terraform-google-project-factory/tree/master/modules/svpc_service_project. If you can open an issue, we can discuss adding it there.

PR #732, opened 2022-07-07
feat: updated project_id output

apeabody:


comment-bot-dev:
@apeabody

bharathkkb (commented):
Whats the reason for this change @apeabody 
I think we historically had `module.project_services.project_id` as output depends_on was not respected.

PR #766, opened 2022-11-10
Configure propagation time (time sleep) for inclusion of the created project in a VPC-SC perimeter

daniel-cit:
Configure the propagation time (time sleep) to be used between the inclusion of the project create in the VPC-SC perimeter and the inclusion of the project create in the shared VPC.

This change is needed  if the API `compute.googleapis.com` is in the list of restricted services of the VPC-SC perimeter

### DIFF
--- a/modules/core_project_factory/main.tf
+++ b/modules/core_project_factory/main.tf
@@ -108,10 +108,10 @@ module "project_services" {
 /******************************************
   Shared VPC configuration
  *****************************************/
-resource "time_sleep" "wait_5_seconds" {
+resource "time_sleep" "wait_vpc_sc_propagation" {
   count           = var.vpc_service_control_attach_enabled ? 1 : 0
   depends_on      = [google_access_context_manager_service_perimeter_resource.service_perimeter_attachment[0], google_project_service.enable_access_context_manager[0]]
-  create_duration = "5s"
+  create_duration = var.vpc_service_control_sleep_duration
 }
 
 resource "google_compute_shared_vpc_service_project" "shared_vpc_attachment" {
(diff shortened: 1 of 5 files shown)


comment-bot-dev:
@daniel-cit

modules/core_project_factory/main.tf
```diff
@@ -108,10 +108,10 @@ module "project_services" {
 /******************************************
   Shared VPC configuration
  *****************************************/
-resource "time_sleep" "wait_5_seconds" {
+resource "time_sleep" "wait_vpc_sc_propagation" {
```
bharathkkb, line 111: @daniel-cit can we keep the same resource name so we can release this in a minor release?
We can add a comment to clean up for next breaking change.

variables.tf
```diff
@@ -299,6 +299,12 @@ variable "vpc_service_control_perimeter_name" {
   default     = null
 }
 
+variable "vpc_service_control_sleep_duration" {
+  description = "The duration to sleep in seconds before adding the project to a shared VPC after the project is added to the VPC Service Control Perimeter"
```
bharathkkb, line 303:
```suggestion
  description = "The duration to sleep in seconds before adding the project to a shared VPC after the project is added to the VPC Service Control Perimeter. VPC-SC is eventually consistent."
```

bharathkkb (commented):
Thanks @daniel-cit

comment-bot-dev:
@daniel-cit

modules/core_project_factory/main.tf
daniel-cit: resource rename change reverted

variables.tf
daniel-cit: Description updated

comment-bot-dev:
@daniel-cit

bharathkkb (approved):

Issue #780, opened 2023-01-05
converge vpc-sc-project-local test is suddently failing due to lack of project_id


apeabody:
### TL;DR

Config: https://github.com/terraform-google-modules/terraform-google-project-factory/blob/7a40ef284c070dcff4c92d75dba8e31a1b88dcec/build/int.cloudbuild.yaml#L130

Example: https://pantheon.corp.google.com/cloud-build/builds/2cf64e84-f62f-4c99-a0b1-205af7bbd058;step=21?project=cloud-foundation-cicd

### Terraform Configuration

```hcl
CI
```

### Terraform Version

```sh
CI
```

bharathkkb:
It might have been https://github.com/terraform-google-modules/terraform-google-vpc-service-controls/issues/103

PR #791, opened 2023-03-02
fix: removes policy id env var from int build config

g-awmalik:
This is a partial rollback of: #788.

The reason #780 was failing was due to lack of permissions for the cloudbuild service agent on the test org. Giving that SA access context reader permission [fixes](https://github.com/GoogleCloudPlatform/cloud-foundation-toolkit/pull/1420) this.


g-awmalik:
cc @imrannayer

apeabody (approved):

PR #799, opened 2023-03-14
feat: added metastore support in shared_vpc_access module

DrFaust92:


DrFaust92:
ping [morgante](https://github.com/morgante) if you can help with this

PR #822, opened 2023-06-12
fix: add service account access for gke in shared vpc

tenminus11:
According to https://cloud.google.com/kubernetes-engine/docs/how-to/cluster-shared-vpc#enabling_and_granting_roles

roles/compute.networkUser access on SERVICE_PROJECT_NUM@cloudservices.gserviceaccount.com in subnetwork is missing in the terraform module

current terraform module fails to deploy a usable cluster, and system related pods doesn't come up
This fix give the required access which is needed


bharathkkb (commented):
Thanks for the PR @tenminus11 

IIRC we actually grant these in the main module. Are you using a different module to create projects?

https://github.com/terraform-google-modules/terraform-google-project-factory/blob/6b0412e61a564477128a3fd2e0735e00aae5e572/modules/core_project_factory/main.tf#L250-L272

bharathkkb:
Discussed offline, usecase was for just svpc setup without project creation. This should be no-op if used together with main module.

bharathkkb (approved):

PR #829, opened 2023-07-19
feat(budget): Only read project data when budget will be created

gleichda:
When having a lot of projects the data resource makes the plan hard to read. To fix that only read the project data if it is needed.


bharathkkb (approved):
Thanks for the PR @gleichda. I think the lint is failing due to an unrelated issue, could you run `make docker_generate_docs` to sync them?

Issue #831, opened 2023-08-18
Datastream default SA doesn't get granted required shared VPC role


dtsong:
### TL;DR

In the current module implementation, the Datastream default service account is granted `compute.networkUser` which is insufficient. due to this variable being defaulted to `false` here: https://github.com/terraform-google-modules/terraform-google-project-factory/blob/master/modules/shared_vpc_access/variables.tf#L65

<img width="826" alt="Screen Shot 2023-08-18 at 4 02 52 PM" src="https://github.com/terraform-google-modules/terraform-google-project-factory/assets/6956872/0d835f2c-1726-4288-a360-ef718de1cb94">

### Expected behavior

The default Datastream service account is granted `roles/compute.networkAdmin` in the shared VPC project so that it could allocate the provided IP range.

### Observed behavior

The default Datastream service account is granted `roles/compute.networkUser` in the shared VPC project which led to a permissions error:

```
 Error: Error creating PrivateConnection: googleapi: Error 400: Missing permissions to validate that the provided IP range is available. Please check that Datastream SA has the required permissions.

com.google.apps.framework.request.StatusException: <eye3 title='FAILED_PRECONDITION'/> generic::FAILED_PRECONDITION: Missing permissions to validate that the provided IP range is available. Please check that Datastream SA has the required permissions.
```

### Terraform Configuration

```hcl
module "project-factory" {
  source  = "terraform-google-modules/project-factory/google"
  version = "~> 14.2"

  name                 = var.project_name
  org_id               = var.org_id
  billing_account      = var.billing_account
  svpc_host_project_id = var.shared_vpc_host_project_id
  shared_vpc_subnets   = var.shared_vpc_subnets
  activate_apis        = var.service_apis
}

```

# terraform.tfvars
```
billing_account            = "your-billing-account"
org_id                     = "your-org-id"
project_name               = "your-new-project-id"
shared_vpc_host_project_id = "your-shared-vpc-project-id"
shared_vpc_subnets         = ["name-of-subnet"]
service_apis = [
    "datastream.googleapis.com" # We have other APIs, but just including Datastream
  ]
```

### Terraform Version
```
Terraform v1.5.5
on darwin_arm64
+ provider registry.terraform.io/hashicorp/google v4.78.0
+ provider registry.terraform.io/hashicorp/google-beta v4.78.0
+ provider registry.terraform.io/hashicorp/null v3.2.1
[...]

PR #833, opened 2023-08-27
fix: grant compute.networkAdmin for Datastream SA by default

dtsong:
Intend to fix: https://github.com/terraform-google-modules/terraform-google-project-factory/issues/831

Per the docs, in a shared VPC context, the host project will need to grant `compute.networkAdmin` to the child project's Datastream service account to support private connectivity.

Screen Shot 2023-08-26 at 7 24 20 PM

Source: https://cloud.google.com/datastream/docs/create-a-private-connectivity-configuration#before-you-begin


imrannayer:
@dtsong we dont assign admin permissions by default. Maybe we can add in variable definition that this is required for datastream.

PR #843, opened 2023-10-05
feat: add support for tpg v5

apeabody:


examples/app_engine/versions.tf
```diff
@@ -18,12 +18,10 @@ terraform {
   required_version = ">= 0.13"
   required_providers {
```
bharathkkb, line 19: Are we keeping this so as not to break incase of null/random provider major updates?

versions.tf
```diff
@@ -18,12 +18,10 @@ terraform {
   required_version = ">=0.13.0"
   required_providers {
     google = {
-      source  = "hashicorp/google"
-      version = "~> 4.28"
```
bharathkkb, line 22: Since this is a module versions let's keep the <6.0?

bharathkkb (approved):
Thanks @apeabody!

versions.tf
apeabody: Yup, good catch, probably a bad find criteria.

examples/app_engine/versions.tf
apeabody: I only pulled the google, google-beta, let me expand to the rest in examples.

bharathkkb: IMO we could just remove versions.tf from examples

apeabody: Yeah - I will try that.

apeabody: Two examples use a provider that is not required by any modules, so need to keep those, but we can remove the rest.

PR #856, opened 2023-11-10
fix: shared_vpc_access - Grant notebooks.googleapi.com SA the networkUser role

derhally:
Update the shared_vpc_access module to grant the notebooks.googleapis.com SA the appropriate network roles. It needs roles/compute.networkUser on the subnets being shared.


bharathkkb (approved):
Thanks for the PR @derhally

PR #874, opened 2024-01-03
fix: shared_vpc_access - Grant workstations.googleapi.com SA the networkUser role

derhally:
Update the shared_vpc_access module to grant the workstations.googleapis.com SA the appropriate network roles. It needs roles/compute.networkUser on the subnets being shared.


derhally:
This is actually not correct.  We had multiple updates going on and it seems like the workstations SA needs networkUser role on the whole project, not just the subnet

PR #885, opened 2024-02-21
feat: support optional tag binding

### LINKED ISSUE #884: Add support for binding a tag with project
### TL;DR

Using google_tags_tag_binding 

### Terraform Resources

_No response_

### Detailed design

_No response_

### Additional information

_No response_

bharathkkb:
fixes #884

### DIFF
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -169,7 +169,7 @@ steps:
   args: ['/bin/bash', '-c', 'cft test run TestQuotaProject --stage destroy --verbose']
 - id: apply-tags-project-example
   waitFor:
-    - init-all
+    - create-all
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestTagsProject --stage apply --verbose']
 - id: verify-tags-project-example
--- a/build/int.cloudbuild.yaml
+++ b/build/int.cloudbuild.yaml
@@ -167,6 +167,26 @@ steps:
     - verify-quota-project-example
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestQuotaProject --stage destroy --verbose']
+- id: init-tags-project-example
+  waitFor:
+    - create-all
+  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
+  args: ['/bin/bash', '-c', 'cft test run TestTagsProject --stage init --verbose']
+- id: apply-tags-project-example
+  waitFor:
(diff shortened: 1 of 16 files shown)


modules/core_project_factory/versions.tf
```diff
@@ -20,11 +20,11 @@ terraform {
   required_providers {
     google = {
       source  = "hashicorp/google"
-      version = ">= 3.50, < 6"
+      version = ">= 3.64, < 6"
```
bharathkkb, line 23: not marking as breaking since this is Internal module and modules using this module are already 4+

modules/core_project_factory/variables.tf
```diff
@@ -258,3 +258,9 @@ variable "grant_network_role" {
   type        = bool
   default     = true
 }
+
+variable "tag_binding_values" {
+  description = "Tag values to bind the project to."
+  type        = list(string)
```
g-awmalik, line 264: Should this be a map(string) because you can specify the key name on `google_tags_tag_key`?

test/setup/tags.tf
g-awmalik, line 1: Why are you setting up tags as part of the setup?

g-awmalik (commented):
Thanks for taking care of this. My comments center around the lack of usage of `google_tags_tag_key` in the module? AFAIK, the key and the value are paired together for the binding and we should give the user the ability to create both. What do you think?

test/setup/tags.tf
bharathkkb: Ah this was because I wanted use random ID in tags to prevent conflicts. I originally had it within example but got unknown values in for_each error, so moved to setup.

modules/core_project_factory/variables.tf
bharathkkb: I chose a list because we weren't passing in key information. I have addressed the main comment about why not include `google_tags_tag_key` too below.

build/int.cloudbuild.yaml
```diff
@@ -167,6 +167,26 @@ steps:
     - verify-quota-project-example
   name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
   args: ['/bin/bash', '-c', 'cft test run TestQuotaProject --stage destroy --verbose']
+- id: init-tags-project-example
+  waitFor:
+    - create-all
+  name: 'gcr.io/cloud-foundation-cicd/$_DOCKER_IMAGE_DEVELOPER_TOOLS:$_DOCKER_TAG_VERSION_DEVELOPER_TOOLS'
+  args: ['/bin/bash', '-c', 'cft test run TestTagsProject --stage init --verbose']
```
apeabody, line 174: This should be safe with bpt v0.12+, however do you still want to init all at the start?

examples/tags_project/variables.tf
```diff
@@ -0,0 +1,35 @@
...
+
+variable "organization_id" {
+  description = "The organization id for the associated services"
+  default     = "684124036889"
+}
+
+variable "folder_id" {
+  description = "The ID of a folder to host this project."
+  type        = string
+  default     = ""
```
apeabody, line 25: Would `null` work/be better?

```suggestion
  default     = null
```

apeabody (approved):
A few minor suggestions - saw the existing comment one why this doesn't need to be a breaking change.

build/int.cloudbuild.yaml
bharathkkb: I removed the single init for now

PR #909, opened 2024-05-10
feat: vpc user to run/functions + soft deletion policy

diegolnasc:


modules/core_project_factory/main.tf
```diff
@@ -310,6 +310,13 @@ resource "google_storage_bucket" "project_bucket" {
   uniform_bucket_level_access = var.bucket_ula
   public_access_prevention    = var.bucket_pap
 
+  dynamic "soft_delete_policy" {
+    for_each = var.soft_delete_policy == {} ? [] : [var.soft_delete_policy]
```
apeabody, line 314: Suggestion only: might be clearer to have the "enabled" outcome first, but checking `!= {}`

apeabody (changes requested):
Thanks for the contribution @diegolnasc!

Looks like we just need test coverage.  Can you please add to one of the /examples, and then as appropriate extend it's integration test in test/integration

PR #921, opened 2024-06-12
feat(TPG>=5.33)!: add support for setting cloud armor tier of the project

imrannayer:


imrannayer:
@g-awmalik plz review

modules/cloud_armor_tier/main.tf
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/******************************************
+  Cloud Armor tier of the project
+ *****************************************/
+
+resource "google_compute_project_cloud_armor_tier" "cloud_armor_tier_config" {
```
bharathkkb, line 21: Since it's a single resource couldn't we keep it within core_projectfactory? I don't see much potential for reuse as it is a single resource module.

imrannayer: Thats what I was thinking. But added as sub-module based on other sub-modules. Leme add it as part of core_projectfactory

modules/core_project_factory/outputs.tf
```diff
@@ -105,3 +105,8 @@ output "usage_report_export_bucket" {
   description = "GCE usage reports bucket"
   value       = try(google_project_usage_export_bucket.usage_report_export[0], null)
 }
+
+output "cloud_armor_tier" {
+  description = "Cloud Armor tier for the project"
+  value       = var.cloud_armor_tier
+}
```
bharathkkb, line 112: I don't think there is much benefit redirect inputs as output so let's remove this

modules/core_project_factory/variables.tf
```diff
@@ -270,3 +270,9 @@ variable "tag_binding_values" {
   type        = list(string)
   default     = []
 }
+
+variable "cloud_armor_tier" {
```
bharathkkb, line 274: Shouldn't we also expose this within the main module?

PR #922, opened 2024-06-13
fix: add missing readme files and regenerate metadata

gajdaw:


modules/core_project_factory/README.md
```diff
@@ -0,0 +1,74 @@
+# Core Project Factory
```
bharathkkb, line 1: These readme's are absent by design as these are considered internal modules. Let's remove any new readme and metadata added here and in the other two places.

bharathkkb (changes requested):
Thanks for the PR @gajdaw

PR #983, opened 2025-03-19
fix: add possible values for "deletion_policy"

muffl0n:


apeabody:
switched to `fix` as this changes a `tf` file

PR #992, opened 2025-05-30
fix(docs): correct typo in GKE shared VPC example readme

pdbrito:


apeabody (approved):
Thanks for the contribution @pdbrito!

PR #994, opened 2025-06-23
feat(core_project_factor): enable setting a description for the default service account

FalconerTC:
Enable setting a description for the default service account in `core_project_factory`. I am implementing an organization policy that requires our SA's have descriptions and we rely on this SA for new projects, so this minor feature request would help us there. The default is null to prevent changing existing resources.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,7 @@ determining that location is as follows:
 | name | The name for the project | `string` | n/a | yes |
 | org\_id | The organization ID. | `string` | `null` | no |
 | project\_id | The ID to give the project. If not provided, the `name` will be used. | `string` | `""` | no |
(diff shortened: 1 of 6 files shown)


apeabody (changes requested):
Thanks for the contribution @FalconerTC!

Please run the `make docker_generate_docs` command and commit the above changes:
```
Checking for documentation generation
diff -r '--exclude=.terraform' '--exclude=.kitchen' '--exclude=autogen' '--exclude=*.tfvars' '--exclude=*metadata.yaml' /workspace/README.md /tmp/tmp.Lb8Vou0Hk7/generate_docs/workspace/README.md
156a157
> | project\_sa\_description | Description to set for the project default service account. | `string` | `null` | no |
Error: Documentation generation has not been run, please run the
'make docker_generate_docs' command and commit the above changes.
```

PR #1015, opened 2025-11-06
Avoid unnecessary budget changes

gorge511:
Hi, we noticed that almost any change in project settings or change on dependent module, like folder in which the project is created, triggers project number refresh via data resource and plan budget update because of that. The budget update is not performed, because at the apply phase the data resource returns identical project number.

```hcl
  # module.my_project.module.budget.data.google_project.project[0] will be read during apply
  # (depends on a resource or a module with changes pending)
 <= data "google_project" "project" {
  ...
      + project_id          = "my-project-1234"
  ...
    }

  # module.my_project.module.budget.google_billing_budget.budget[0] will be updated in-place
  ~ resource "google_billing_budget" "budget" {
  ...
      ~ budget_filter {
          ~ projects               = [
              - "projects/1234567890",
            ] -> (known after apply)
        }
    }

  # module.my_project.module.project-factory.google_service_account.default_service_account[0] will be created
  + resource "google_service_account" "default_service_account" {
      + account_id                   = "project-service-account"
      + create_ignore_already_exists = true
      + disabled                     = false
      + display_name                 = "my-project-1234 Project Service Account"
      + email                        = "project-service-account@my-project-1234.iam.gserviceaccount.com"
      + id                           = (known after apply)
      + member                       = "serviceAccount:project-service-account@my-project-1234.iam.gserviceaccount.com"
      + name                         = (known after apply)
      + project                      = "my-project-1234"
      + unique_id                    = (known after apply)
    }

Plan: 1 to add, 1 to change, 0 to destroy.
```

As the project number is immutable thing and it's known and already available in `core_project_factory` module outputs, I suggest passing it to the budget module directly to avoid using the data resource at all.

To keep the budget module backward compatible and still usable as standalone module, it will still fetch the project numbers when they are not provided directly.

### DIFF
--- a/main.tf
+++ b/main.tf
@@ -102,6 +102,7 @@ module "budget" {
   create_budget = var.budget_amount != null
 
   projects                         = [module.project-factory.project_id]
+  project_numbers                  = [module.project-factory.project_number]
   billing_account                  = var.billing_account
   amount                           = var.budget_amount
   alert_spent_percents             = var.budget_alert_spent_percents
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -382,6 +382,6 @@ spec:
       - serviceusage.googleapis.com
     providerVersions:
       - source: hashicorp/google
-        version: ">= 5.41, < 7"
+        version: ">= 5.41, < 8"
       - source: hashicorp/google-beta
-        version: ">= 5.41, < 7"
+        version: ">= 5.41, < 8"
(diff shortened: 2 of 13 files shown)


gorge511:
This PR is still valid. Just needs to be reviewed.

joaonunatings:
Not sure why this PR is not reviewed, I get the same issue (v18.2).

PR #1025, opened 2026-01-01
Update main.tf

PaggieChen:
fix(iam): resolve null service identity email error

Refactor service agent identity logic to use deterministic email construction.

Calculate service agent emails using ${project_number} to bypass API race conditions.

Add explicit depends_on between project services and IAM members.

Standardize Compute Engine default service account retrieval.


regressEdo:
what's the status here?

PR #1034, opened 2026-04-24
feat: make project lien reason customizable

mmajor-rtbh:
Add lien_reason variable to allow customization of the project lien message.

The variable defaults to empty string and falls back to "Project Factory lien" when not provided, maintaining backward compatibility.


mmajor-rtbh:
bump

colinpusey:
This would be super useful for us as well

mmajor-rtbh:
> This would be super useful for us as well

I'am waiting for it too. but idk if this repo is still alive or abadoned

